Issue #4, opened 2021-06-17
Managing key secret pair in a yaml file rather than plainly in .py files


raphaeleadjerad:
- create a `secrets.yaml` file
- add it to .gitignore
- tell people to put their key-secret pair in it
- all code using the pair should load the yaml instead of writing it, as follows:

```python
import yaml
with open("secrets.yaml", "r") as creds:
        secrets = yaml.safe_load(creds) 
```

hadrilec:
Done  #12

PR #12, opened 2021-06-17
[WIP] : add secrets to a yaml and safe_load

raphaeleadjerad:
WIP on #4
Do not merge yet, should have created a draft pr instead

Should in fact rather create a function `init_conn` rather than these lines


hadrilec:
Creation of the init_conn function, it stores the credentials in a yaml file saved in the HOME directory.
The yaml file is then automatically read before any request.
Done.

PR #19, opened 2021-07-12
[WIP] Adds a new module download to get files that are not accessible using API

### LINKED ISSUE #17: Additional module to directly download files from Insee website
Many Insee's dataset are not yet available in API. I think we can recycle lots of stuff from [doremifasol R package](https://github.com/InseeFrLab/DoReMIFaSol) to propose a simple approach in `python`.  This could be an additional module in `pyinsee`, let's say `pyinsee.download`.

Within `doremifasol`, downloads are controlled from [a json file](https://raw.githubusercontent.com/InseeFrLab/DoReMIFaSol/master/data-raw/liste_donnees.json). We could directly use this file. This would make that module easier to maintain because both `doremifasol` and `pyinsee.download`  would rely on that. 

@hadrilec I am volunteer to explore this approach if you think this is an interesting one

linogaliana:
This module aims to propose a new framework to get data from insee.fr . The approach will be based on `request` from a registry of data sources [there](https://raw.githubusercontent.com/InseeFrLab/DoReMIFaSol/master/data-raw/liste_donnees.json)

⚠️ Work in Progress

close #17 

# Progress

Steps to reproduce [doremifasol framework](https://github.com/InseeFrLab/DoReMIFaSol) :  

* [x] importing JSON
* [x] telechargerFichier
* [x] chargerDonnees 
* [x] telechargerDonnees
* [x] unit tests
* [ ] rest API

Format implemented :

* [x] zip
* [x] csv
* [x] xls/xlsx
* [ ] JSON

Improvements to consider :

* [x] progress bar for `requests`


linogaliana:
Une difficulté : les clés du JSON ne sont pas uniques. La solution à l'heure actuelle est pas 100% fiable

Issue #38, opened 2021-11-12
fix tests for better ci (quality)


cthiounn:
pynsee's pipeline are in errors since august. Best practices advise to keep it working in order to detect bugs and regressing code. 

It seems there are two errors, one minor and one bigger : 

minor : a small typo indentation in test_pynsee_sirene.py
complex : the test of the download of the census data is not working, because pandas in the github runner cannot achieve to load the 4gb CSV file

linogaliana:
- Thanks for digging into our CI problems. Yes you are right with the RP problem. Given the current implementation of `telechargerDonnees` (name will be changed, see #37) it is not trivial. A workaround could be to add a `nrow` argument to the function that would be passed to `pandas` to avoid trying read the entire file

linogaliana:
However, maybe that would make more sense not to test `telechargerDonnees` with RP. We could just check the `requests` 200 code to be sure we could find the dataframe if needed

PR #39, opened 2021-11-12
Fix tests

cthiounn:
#38 Resolve the two errors in order to fix the test pipeline

Beyond that, the census data download functionality raise the question of size file for test purpose (each pipeline are downloading 300mo *4) , the scope of the test itself (test the function or the corresponding use case?) and the running speed of tests suite. Test can be parallelized with pytest-xdist. I didn't test it thoroughly, then I didn't submit that implementation, and it could be a source of complexity later, but feel free to test it and implement it, to review it the other dev.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -24,7 +24,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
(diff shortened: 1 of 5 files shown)


linogaliana:
Hi Conrad,
Thanks for your PR. I am going to integrate your suggestions into another branch that brings substantial changes to the `download` module.

linogaliana:
Regarding the download of files, I do agree, this is not a good strategy to integrate that into tests. I will probably change that whenever possible

PR #41, opened 2022-01-20
new module geodata

hadrilec:
download geographical data from IGN APIs


hadrilec:
examples are fine so the new module is ok

PR #46, opened 2022-02-07
Devfix - json instead of pickle + bug fix on get_location method of sirenedataframe

hadrilec:


hadrilec:
memory error 127 - json files too big - pickle files are kept instead

PR #50, opened 2022-02-12
Translate overseas + zoom_paris - full map

hadrilec:


hadrilec:
outdated - new functions coming soon from another branch

PR #65, opened 2022-03-21
Stringdist - remove levenshtein as dependency 

hadrilec:
Hi @linogaliana,
I have slightly modified the way suggestions are made to remove python-levenshtein package from dependencies as it is troublesome to install on Windows.
Is this fine with you?

see commit 963170190fd3762d7e94477aa8df445848716612

A+
Hadrien


linogaliana:
Great if we can avoid this troublesome dependency

linogaliana (commented):
changes are fine for me

PR #75, opened 2022-04-24
Publish - release package after build

hadrilec:


hadrilec:
a tag should be attributed to one commit on the master branch:
git tag v0.0.999
then it should be push to remote:
git push origin v0.0.999
the wheel and tar.gz will be built and then pushed to githup repo and testpypi

PR #116, opened 2023-01-18
Correct error when providing a string for http_proxy

tfardet:
np.isnan does not support str entries so the test used to trigger an error whenever an actual value was provided

### DIFF
--- a/pynsee/utils/_get_credentials.py
+++ b/pynsee/utils/_get_credentials.py
@@ -18,7 +18,7 @@ def _get_credentials():
         os.environ["insee_key"] = str(cred.loc[0, "insee_key"])
         os.environ["insee_secret"] = str(cred.loc[0, "insee_secret"])
         proxy_server = cred.loc[0, "proxy_server"]
-        if (proxy_server is None) or (np.isnan(proxy_server)):
+        if (proxy_server is None) or (not isinstance(proxy_server, str)):
             proxy_server = ""
         os.environ["http_proxy"] = str(proxy_server)
         os.environ["https_proxy"] = str(proxy_server)


tfardet:
Don't know if I'm supposed to do something for the tests to pass... clearly some credentials are missing but I guess it should be configured to use your secrets for PRs, right?
It runs fine on my local machine with my credentials.

hadrilec:
Indeed I have to check how to allow external PRs to have access to repo's secrets, if you find any guideline on this let me know, in any case I will check how to do it in the coming days

hadrilec:
I have just added you as an active contributor to this repo, let me know if it goes in the right direction

tfardet:
Thanks, I'll try to have a look (hopefully this evening or tomorrow), I'm not sure that write access alone will change the problem, though, probably there'll be something more to do.

tfardet:
OK, I looked around and I think ``pull_request_target`` is the way to go!
See [here](https://github.com/actions/checkout/issues/518) for a discussion about the actual code that should be used:

```yaml
name: pynsee package tests

on:
  push:
      branches:
        - 'master'
  pull_request_target:
    branches:
        - 'master'

jobs:
  build:
    runs-on: ubuntu-latest
    strategy:
      matrix:
        python-version: ["3.7", "3.8", "3.9", "3.10"]

    steps:
      - name: Checkout code
        uses: actions/checkout@v2
        with:
          ref: "refs/pull/${{ github.event.number }}/merge"
```

and [this page](https://dev.to/petrsvihlik/using-environment-protection-rules-to-secure-secrets-when-building-external-forks-with-pullrequesttarget-hci) for a discussion about how to make it safe.

hadrilec:
@tfardet thanks a lot, I modified the package test yaml file following your guidelines.
I also included your bug fix in this commit d508b1283d1b03fe2fa42e9c64efb398a82f9faa
If you are working behind a proxy server, I would be happy if you could let me know if pynsee is still usable behind the proxy

Issue #118, opened 2023-02-01
get_new_city - error message not relevant when uptodate city


tgrandje:
When you try the function with an uptodate city, you get a false error message :
```
import pynsee
r = pynsee.localdata.get_new_city('59350')
```
Return was :
```
!!! An error occurred !!!
Query : https://api.insee.fr/metadonnees/V1/geo/commune/59350/suivants?date=2013-01-01


!!! Make sure you have subscribed to all APIs !!!
Click on all APIs' icons one by one, select your application, and click on Subscribe

Error 404

Traceback (most recent call last):

  Cell In[13], line 1
    r = pynsee.localdata.get_new_city('59350')

  File ~\AppData\Local\pypoetry\Cache\virtualenvs\sis-scraping-0fCjv732-py3.9\lib\site-packages\pynsee\localdata\get_new_city.py:50 in get_new_city
    request = _request_insee(api_url=api_link, file_format="application/json")

  File ~\AppData\Local\pypoetry\Cache\virtualenvs\sis-scraping-0fCjv732-py3.9\lib\site-packages\pynsee\utils\_request_insee.py:116 in _request_insee
    raise ValueError(results.text)

ValueError
```

The API seems to return 404 indeed (which imho is strange behaviour, a specific code might have been employed). I'm not sure how to proceed regarding the error message though (which relates to the API subscription and has nothing to do with the 'problem' here).

hadrilec:
Indeed I have to improve the error messages to make them more self explanatory.
In this case, I guess the data does not exist but I will double check later this week

tgrandje:
Yes, that's it : the API returns 404 ("Ressource non trouvée") if the data does not exist. The authentication problem is supposed to be a 401 (that is, on the geographical API at least). Are you ok with me I trying to fix this in _request_insee.py, or do you want to have a check on all other APIs beforehand ?

hadrilec:
go ahead with a PR it is more than welcome because I have a lot of stuff on my workbench already.
I added this error about the credentials because the API does not really return it and as it is not easy to understand how api.insee.fr works I wanted to make sure to raise the user's attention on this particular point. So, maybe the wording is not crystal clear, but the content should be reflected anyway in the error message because it is important in some cases

PR #121, opened 2023-02-02
Detail exceptions based on APIs' return codes

tgrandje:
Added detailed responses about the APIs responses codes.

I checked the documentation of all current APIs, but there may be some return codes which have not been documented in the swagger. For instance, the SIRENE API's doc mentions 301 codes for permanently moved companies ; I haven't seen those responses while scrolling through the swagger (and I haven't checked if requests handles the redirection by itself as I have no concrete example to test it).

I changed the class of errors returned when encountering the 40X & 50X errors. I think 'ValueError' was an awkward choice (regarding to the large explanations about those errors). I choose instead a generic "requests.exceptions.RequestException which is more neutral (I think...).

(I maintained the previous error messages/classes in the case of the APIs returning any other code than those expressly stated.)


tgrandje:
Linked to https://github.com/InseeFrLab/pynsee/issues/118

PR #130, opened 2023-03-13
Shapely>=2.0 : upgrade usable shapely version 

hadrilec:


Quentin-Guignard-ED:
Thanks for you work !
Do you think you could merge this feature ?

There is a conflict to use PYNSEE and OSMNx packages cause to a conflict on shapely version. (The second one needs shapely >=2.0)

hadrilec:
#134 contains the commits

PR #129, opened 2023-03-08
feat: use logging to replace print #117

lgeo3:
First draft to use logging to replace print (issue #117)

PR #137, opened 2023-05-10
Patch cheflieu

tgrandje:
Keep "CHEFLIEU" generated by INSEE's API when working on départements & arrondissements.

### DIFF
--- a/pynsee/__init__.py
+++ b/pynsee/__init__.py
@@ -1,3 +1,6 @@
+import logging
+import os
+
 from pynsee.sirene import *
 from pynsee.macrodata import *
 from pynsee.localdata import *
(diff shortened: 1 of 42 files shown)


hadrilec:
@tgrandje I would like to push on your branch, I did not manage, could you please double check whether I have the right to do so?
Thanks

hadrilec:
I have tried to run a couple of notebooks and I dont see anything printed.

I am fine with having the option to disable info messages from pynsee, but I believe that by default they should be printed that they provide key information about how the package works.

Could you please make sure messages are by default printed (including in jupyter notebooks)?

thanks a lot

tgrandje:
I just added you to the fork's collaborators (sorry about that, as I created a fork of @lgeo3's own fork I lost you in the process).

tgrandje:
Regarding to the display of messages: I certainly can set the default logging's display threshold to info. But it is considered bad practice afaik (as I already stated in https://github.com/InseeFrLab/pynsee/pull/129 ).

What I tried to do (and this is clearly something that needs to be reviewed by somebody else), was :
* to display as info what is relevant to the normal workflow (to cite @tfardet 's own words). For instance, if the user set tokens through environment variables, I think it safe to assume he did it on purpose and won't need a warning about it;
* to display "upper level" logger entries (ie. warning, error or critical) when something odd appeared during the process.

So, to sum it up, I'd say we should review the levels collectively and not set the logging's level globally.

In the meantime, a practical hack for your notebook could be to configure the root logger's level beforehand (for instance in the notebook's second cell, just after calling the modules). For instance, something like this should work :
```
import logging
logging.basicConfig(level=logging.INFO)
```

tgrandje:
@hadrilec are your notebooks in the current repo ?

hadrilec:
Yes all notebooks are in the repo

Issue #140, opened 2023-05-11
pynsee.localdata.get_geo_list on arrondissements descendants downloads communes


tgrandje:
As the "type" is not passed to _get_geo_relation when working on "Arrondissements", all descendants are downloaded from the API, resulting in heavy traffic. It should also be the case regarding other available territories'types ('CommunesDeleguees' and so on) : I will check that.

hadrilec:
could you confirm that #142 solved the issue?

tgrandje:
It should, though I'm not entirely convinced the try/except was the best of ideas anymore. It seems that "most of the time", you'll get fastest results this way ; on the other hand it can also consume twice the api's requests' level when it fails. I'll try to benchmark that in details next week to be more thorough (and maybe rollback that change).

Issue #139, opened 2023-05-11
pynsee.localdata.get_geo_list throws an exception when encoutering obsolete territories


tgrandje:
The "DateSuppression" field returned by the API in that case will trigger an exception (as the number of columns in the generated dataframe will change).

```
from pynsee.localdata import get_geo_list
df = get_geo_list("departements", "1990-01-01")
print(df.query('CODE=="01"'))
```

hadrilec:
solved by #142

Issue #138, opened 2023-05-11
pynsee.get_geo_list not concordant with given date for regions


tgrandje:
get_geo_list's date argument is ignored when querying on regions.

```
from pynsee.localdata import get_geo_list
df = get_geo_list("regions", "1990-01-01")
print(df.query('CODE=="32"'))
```

Issue #136, opened 2023-05-10
pynsee.localdata.get_geo_list doesn't keep "CHEFLIEU" columns if geo is not "region"


tgrandje:
Hello Hadrien
I see that get_geo_list returns a dataframe with "CHEFLIEU" as a column, but that is not the case if you ask for (say) a departement.
I've tracked it down to [line 172](https://github.com/InseeFrLab/pynsee/blob/master/pynsee/localdata/get_geo_list.py#L172) and that seems intentional.
Do you have objections to me setting a PR regarding that behaviour ?
Thanks

hadrilec:
@tgrandje ok fine, yes please make a PR I will review it whenever I have a bit of time, thanks a lot

tgrandje:
https://github.com/InseeFrLab/pynsee/pull/137

tgrandje:
Just done my PR (which will also includes the PR about the logger, sorry about that...).

hadrilec:
solved by #142

PR #141, opened 2023-05-12
Logging / get_geo_list diverse fixes

tgrandje:
I haved replaced the previous PR. This should allow :

- to set the logging level beforehand (there was a glitch before) ;
- fix issues on get_geo_list (https://github.com/InseeFrLab/pynsee/issues/140 https://github.com/InseeFrLab/pynsee/issues/139 https://github.com/InseeFrLab/pynsee/issues/138 and https://github.com/InseeFrLab/pynsee/issues/136)

### DIFF
--- a/pynsee/localdata/_get_geo_list_simple.py
+++ b/pynsee/localdata/_get_geo_list_simple.py
@@ -12,9 +12,10 @@ from pynsee.utils._get_temp_dir import _get_temp_dir
 
 
 @lru_cache(maxsize=None)
-def _get_geo_list_simple(geo, progress_bar=False):
-
+def _get_geo_list_simple(geo, date=None, progress_bar=False):
     api_url = "https://api.insee.fr/metadonnees/V1/geo/" + geo
+    if date:
+        api_url += f"?date={date}"
     results = _request_insee(api_url=api_url, sdmx_url=None)
 
     dirpath = _get_temp_dir()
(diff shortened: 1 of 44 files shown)


tgrandje:
Note that the logging is actually totally standard. You can active a "verbose" mode by using the following syntax :
```
import logging
logging.basicConfig(level=logging.DEBUG)
from pynsee.localdata import get_geo_list
df = get_geo_list("regions", "2023-01-01", update=True)
```

pynsee/localdata/_get_geo_list_simple.py
```diff
@@ -12,9 +12,10 @@
 
 
 @lru_cache(maxsize=None)
-def _get_geo_list_simple(geo, progress_bar=False):
-
+def _get_geo_list_simple(geo, date=None, progress_bar=False):
```
tfardet, line 15: shouldn't this be in a separate PR? this one is already quite big (for intrinsic reasons) and this change is drowned in unrelated changes.

pynsee/localdata/get_geo_list.py
```diff
@@ -19,6 +24,8 @@ def get_geo_list(geo=None, date=None, update=False):
     Args:
         geo (str): choose among : communes, communesDeleguees, communesAssociees, regions, departements, arrondissements, arrondissementsMunicipaux
 
+        date (str): date of validity (AAAA-MM-DD)
```
tfardet, line 27: same remark

tfardet (changes requested):
Thanks for opening a new PR for logging since the PR seemed stalled.
I would recommend to split the ``get_geo_list`` part out into a new PR. Mixing both makes it impractical to see and discuss the latter.

PR #144, opened 2023-05-15
Fix bugs on get_ascending_area & get_descending_area

tgrandje:
* fix bug on url construction (if date is missing, "?" was missing from the URL)
* fix message on ValueError if area not found
* black formatting

### DIFF
--- a/tests/localdata/test_pynsee_localdata.py
+++ b/tests/localdata/test_pynsee_localdata.py
@@ -241,11 +241,13 @@ class TestFunction(TestCase):
             df = get_descending_area("zoneDEmploi2020", code="1109")
             test = test & isinstance(df, pd.DataFrame)
 
-            df = get_descending_area("departement", code="59", update=True)
+            df = get_descending_area(
+                "departement", code="59", type="arrondissement", update=True
+            )
             test = test & isinstance(df, pd.DataFrame)
 
-            df = get_ascending_area(
-                "departement", code="59", type="region", update=False
+            df = get_descending_area(
+                "departement", code="59", type="arrondissement", update=False
             )
             test = test & isinstance(df, pd.DataFrame)
 
--- a/tests/localdata/test_pynsee_localdata.py
+++ b/tests/localdata/test_pynsee_localdata.py
@@ -22,8 +22,8 @@ from pynsee.localdata.get_new_city import get_new_city
 from pynsee.localdata.get_ascending_area import get_ascending_area
 from pynsee.localdata.get_descending_area import get_descending_area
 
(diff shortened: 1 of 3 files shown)


tfardet (changes requested):
Looks good overall, I'm just wondering about the hash function... is there a specific reason why you're not using hash on the URL?
If there is, then it would be better to write a hash function and call it in both functions rather than duplicate code.
Could you also merge the main branch again?

tgrandje:
@tfardet about _hash I did it according to other parts of the module (for instance, it is consistent with [that part](https://github.com/InseeFrLab/pynsee/blob/master/pynsee/localdata/get_geo_list.py#L68)).
(I think that most of the time, the URL is computed afterwards and after trying to access the cached file (for which you need the hashed value). It might be faster that way, but not by much...)

tgrandje:
@tfardet I think there are tests already : [here](https://github.com/InseeFrLab/pynsee/blob/master/tests/localdata/test_pynsee_localdata.py#L204) and [there](https://github.com/InseeFrLab/pynsee/blob/master/tests/localdata/test_pynsee_localdata.py#L195) ?

tfardet:
@tgrandje I understand, it just feels a bit bad to have two additional lines of code (repeated in two different functions) just to prepare a list which has only one purpose: to be used to compute a hash... but I guess it outside the scope of this PR.

Regarding the tests:  it would be good to have one with the ``type`` argument but without the ``date``, so we actually check that this bug does not occur anymore (in case someone make a revert from an old branch or whatever)

hadrilec:
@tfardet indeed if you dont mind I would prefer to make the merges myself, but your reviews are highly appreciated on my side as my review becomes just a routine check. Thanks again @tgrandje and @tfardet, great to have you on board ;)

tests/localdata/test_pynsee_localdata.py
```diff
@@ -299,6 +309,13 @@ def get_area_list_test():
 
             self.assertRaises(ValueError, get_area_list_test)
 
+
+        def test_get_area_list_3(self):
+            def get_area_list_test():
+                get_area_list(area="regions", date="1900-01-01", update=True)
+
+            self.assertRaises(RequestException, get_area_list_test)
```
tfardet:
```suggestion
        with self.assertRaises(RequestException):
            get_area_list(area="regions", date="1900-01-01", update=True)
```

tfardet (changes requested):
@hadrilec I think the last change you made does not do what you actually wanted, can you check my proposed change?

hadrilec (approved):
ok

Issue #145, opened 2023-05-15
pynsee.localdata.get_area_list not working with collectivitesDOutreMer argument


tgrandje:
area is converted to lowercase (in line [57](https://github.com/InseeFrLab/pynsee/blob/master/pynsee/localdata/get_area_list.py#L57)) and thus not found inside list_available_area at [line 66](https://github.com/InseeFrLab/pynsee/blob/master/pynsee/localdata/get_area_list.py#L66).

I suspect the area.lower() was introduced to somehow bypass the user's mistakes on area's case. If that's the case, I'd say we should drop it alltogether (for isntance, it shouldn't work for communesAssociees either).

EDIT : it won't work to remove the area.lower() for "UU2020" if we remove the lowercase.

PR #146, opened 2023-05-15
Fix get_area_list

tgrandje:
Should resolve https://github.com/InseeFrLab/pynsee/issues/145


tgrandje:
Just added some more changes to get_area_list. To sum it up, the PR contains :

* a quick and dirty fix of area's case sensitive problem (which would happen with areas rightfully containing upper-case letters, for instance "collectivitesDOutreMer")
* make get_area_list consistent with get_geo_list about "date" argument : 
  * add the argument to the function 
  * change default behaviour (which until now used a "*" wildcard and so got all available data) to the same as get_geo_list : the default is now consistent with the API's documentation (ie. the current date)
* black formatting

pynsee/localdata/get_area_list.py
```diff
@@ -49,40 +54,49 @@ def get_area_list(area=None, update=False):
...
             list_available_area = [area]
 
-    filename = _hash("".join(['get_area_list'] + list_available_area))
+    date_hash = ""
+    if date is not None:
+        date_hash = date
+
+    filename = _hash(
+        "".join(["get_area_list"] + list_available_area + [date_hash])
+    )
```
tfardet, line 85:
```suggestion
    filename = _hash(
        "".join(["get_area_list"] + list_available_area + [date or ""])
    )

```

tfardet (changes requested):
Thanks for the bugfix!
Could you also add a test for these cases?

@hadrilec I'm not familiar with this INSEE data but I'm a bit surprised by the hardcoded date, is this something that will never change of will we see later entries in the future that would make it worth creating an issue to think about how to properly handle this?
(not relevant for that PR, obviously)

tgrandje:
@tfardet that one should be ok now, though I didn't checked the behaviour of the tests. Those are covered in the automated tests directly runned on github, right ?

pynsee/localdata/get_area_list.py
```diff
@@ -110,15 +121,17 @@ def get_area_list(area=None, update=False):
...
     else:
         try:
             data_all = pd.read_pickle(file_data)
-        except:
+        except Exception:
             os.remove(file_data)
             data_all = get_area_list(area=area, update=True)
         else:
-            logger.info(
+            print(
```
tfardet, line 134: something wrong happened here ^^

tfardet (changes requested):
Almost there: you seem to have undone some of your previous hard work by reintegrating a print...

tfardet (approved):

tgrandje:
Sorry, my bad (I should have postponed it to this morning :-) ). Github desktop displayed some strange things while changing of computer this morning but this seems ok on Github (web).

hadrilec:
@tfardet 

> Thanks for the bugfix! Could you also pull from main branch and add a test for these cases?
> 
> @hadrilec I'm not familiar with this INSEE data but I'm a bit surprised by the hardcoded date, is this something that will never change of will we see later entries in the future that would make it worth creating an issue to think about how to properly handle this? (not relevant for that PR, obviously)

Indeed I agree, hard coded dates are far from being state-of-the-art. It is good to know that on paper local data API is still on a beta version. Then, some data update occurs every 5 five years so it is not a big deal to get some dates hard coded as long as the maintenance can keep up.

hadrilec:
Thanks a lot @tgrandje and @tfardet. Just merged after review.

Issue #155, opened 2023-07-07
Add support for INSEE API  get /geo/{area}/{code}/projetes


tgrandje:
This endpoint uses code, date and dateProjection. It therefore can't use _get_geo_relation.

tgrandje:
@hadrilec can you assign me to this issue?

hadrilec:
as the related PR was merged, I guess this issue should be closed, right?

PR #156, opened 2023-07-07
add projection support

tgrandje:
* add get_area_projection
* add a mention of it in get_new_city to explain the difference (couldn't think of a way to explain the reverse, though...)
* add tests
* fixed/optimized some tests

Should cover https://github.com/InseeFrLab/pynsee/issues/155

### DIFF
--- a/pynsee/localdata/get_area_projection.py
+++ b/pynsee/localdata/get_area_projection.py
@@ -1,0 +1,91 @@
+# -*- coding: utf-8 -*-
+
+import pandas as pd
+import datetime
+from functools import lru_cache
+
+from pynsee.utils._request_insee import _request_insee
+
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+@lru_cache(maxsize=None)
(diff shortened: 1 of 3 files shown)


tgrandje:
@hadrilec can you restart these tests? (In regard to your latests comments [here](https://github.com/InseeFrLab/pynsee/issues/160) ). I don't think I have the rights to restart it on my own.

There should'nt be any errors here as it passed the 3.7 build (and nothing I added to the package should be version-specific).

tfardet:
Tests are hitting the max duration wall, I wonder if it's not due to some issues with the IGN server

tgrandje:
@tfardet Mmh... I'm also starting to think that @linogaliana was right about suggesting using cached data (or maybe mockups) for testing purposes. I think the previous job was aborted somewhere in the macrodata testing process : as I haven't touched this section of the package, I'm suspecting INSEE's API's requests rates might be to blame...

tfardet:
@tgrandje agreed -> #164

Issue #154, opened 2023-06-30
get_local_data with the latest data option on popleg does not work


hadrilec:
```
pop = get_local_data(dataset_version='POPLEGlatest',
                       variables =  'IND_POPLEGALES',
                       nivgeo = 'COM',
                      update=True,
                       geocodes = '75056')

```

hadrilec:
solved by  #157

PR #157, opened 2023-07-08
bug fix - get latest local data

hadrilec:
it should solve this issue #154


hadrilec:
only test on get_last_release failed due to the failure of the rss flow, insee teams are working on it, code from this PR is fine

PR #162, opened 2023-07-18
Configuration for credentials, proxies, and progress

### LINKED ISSUE #152: Improve handling of progress bars
At the moment, ``pynsee`` uses ``tqdm`` for progress bars.
It's a very solid library that does the job perfectly but it does not interact well with... basically anything.
So right now, any code or library using ``pynsee`` internally will have to accept having progress bars popping everywhere.

I think it would be nice to provide better options for handling them. I see to main options:

1. the simpler option is to provide one global parameter to hide progress (setting ``disable=True`` to the [tqdm object](https://tqdm.github.io/docs/tqdm/#__init__))
2. switch to [``rich.progress``](https://rich.readthedocs.io/en/stable/progress.html), which, as far as I know, is the only library that supports nested progress bars; one would then have a global ``progress`` object that people could use to [add other progress bars](https://rich.readthedocs.io/en/stable/progress.html?highlight=add_task#advanced-usage) or simply disable them

tfardet:
Use configuration dictionary:

- credentials and proxies are taken from the config dict
- priority order is environment, then config for both
- ``hide_progress`` option to disable progress bars from ``tqdm``

**Question 1:** where should I document that?

Removed all writing to environment except for the token.
**Question 2:** should I also move it to the config dict to avoid writing to env altogether?

Cleaned up some bare excepts and imports.

Fixes #152


tgrandje:
About Q1: I'd say either [here](https://github.com/InseeFrLab/pynsee/tree/master/docs#installation--api-subscription) or [there](https://github.com/InseeFrLab/pynsee/tree/master/docs#how-to-avoid-proxy-issues-)?

Maybe we should also document the fact that tokens are stored to a CSV on the current machine if not stored through environment variables (I seem to remember I moved to environment variables for this very reason).

About Q2: I like the idea about the `_config` dict! I'm totally ok with dropping the env storage myself, bearing in mind that I'm not in favour of storing uncrypted keys on the disk. Maybe we should add _that_ behaviour in the config dict also.
There are other variables stored in the env also (like `pynsee_file_list`); I didn't check all of those (I just encountered one while checking your code). That would make sense to remove those also, doesn't it? @hadrilec what do you think?

pynsee/utils/init_conn.py
```diff
@@ -61,40 +63,33 @@ def init_conn(insee_key, insee_secret, http_proxy="", https_proxy=""):
...
-        pass
+    token = _get_token(insee_key, insee_secret)
 
-    if token is None:
+    if not token:
         raise ValueError(
-            "!!! Token is missing, please check insee_key and insee_secret are correct !!!"
-        )
-    else:
-        logger.info("Token has been created")
```
tgrandje, line 82: Should we keep a trace in log (I've a project which was repeating those calls and this was helpful). Would be better inside `_get_token_from_insee` instead, though.

tgrandje:
Ok, so the extensive values stored in environ by pynsee are:
* `pynsee_file_list`: this is only used in tests, should we keep it or store it in `_config`?
* `pynsee_use_sdmx`: I'm not sure but this one seems to be a "real" argument stored in environ for conveniance's sake?
* `insee_date_test`: this is only used in tests; should not be stored in _config (but might be linked to the refactorization of tests)
* `pynsee_idbank_file`: I'm not sure this one is used outside tests?
* `pynsee_idbank_loop_url` (and `PYNSEE_IDBANK_LOOP_URL`): this is only used in one function (`_dwn_idbank_files`) and defaults to True. Is this used outside of tests (also not documented, I think) ?
* `pynsee_print_url` : this one is related to a missed `print` in _request_insee and this is it. Should be set to logging info/debug and dropped altogether (this is also used in [a notebook](https://github.com/InseeFrLab/pynsee/blob/master/pynsee/localdata/Untitled.ipynb) which I'm not convinced is where it should be).

hadrilec:
Thanks a lot for the comprehensive list of env variables. They are used mainly as workarounds, I do agree it should be cleaned up, so thanks for raising this topic. Below, my take on the topic for each case:

- `pynsee_file_list`: it can be included in a `config` dictionary
- `pynsee_use_sdmx`: it can be included in a `config` dictionary
- `insee_date_test`: this is used only for testing purposes, if you know a better way to perform the test related to this variable, let me know, otherwise I would prefer to keep it as it is
-  `pynsee_idbank_file`: it can be included in a `config` dictionary
- `pynsee_idbank_loop_url`: this is not really useful, the code snippet using it should be removed and the variable not used anymore
- `pynsee_print_url`: this is used mainly for development to check that the queries created are in line with expectations, it can be also useful for some users, I think it can be included in a `config` dictionary

tfardet:
> I'm totally ok with dropping the env storage myself, bearing in mind that I'm not in favour of storing uncrypted keys on the disk. Maybe we should add _that_ behaviour in the config dict also.

Sure, we could set a variable for that, but in that case it would probably make sense to retain the config over multiple runs and that would require a config file. Though it's probably a good idea, maybe this could be done in a later PR.

In the meantime, we could probably write the credentials in a file with restricted access using ``0o600`` with [an opener](https://stackoverflow.com/a/45368120/5962321).

@hadrilec : noted, I'll do that during the 2nd week of August, when I'm back from holiday

tgrandje:
@hadrilec I'll try to think on something with tests to reduce the need for `config` in a separate branch.

About `pynsee_print_url`, that's precisely why we introduced the `logging` feature in the first place : in the last commit I did, you just have to switch the logging level to get the "print"s. If you really want to I'll revert the commit, but I don't think this is needed.

hadrilec:
@tgrandje ok, I am fine for having the queries printed in the log but it should not go under INFO but under DEBUG, otherwise the user will be overwhelmed. Consequently, `pynsee_print_url` should be removed.

tfardet:
I'm wondering about the [``_get_envir_token``](https://github.com/InseeFrLab/pynsee/blob/master/pynsee/utils/_get_envir_token.py#L10) function: is it really necessary to have this test of the token (this is going to be repeated a lot)?
Can't we just work based on the error code from the requests if they fail?

tgrandje:
I'm not sure this is really slowing things (when not working from cache), though I'm saying this on a hunch and haven't time to check. In my experience, the major impact to speed is the API's rate (30requests/minute if I'm not mistaken) and there is not much to do about it...

But maybe we could store a session through the dict config altogether ?

tfardet:
Storing a session in the dict is definitely a great idea, will do!

Regarding the token, besides the time it might take, it's also more of a "why is this needed?" question: the API is going to reply with a 401 unauthorized code if we provide incorrect headers (token) so I don't see the point of checking the token.

tfardet:
@hadrilec just to clarify something: I still plan to allow some configs to be provided via environment variables but in the list that @tgrandje provided, I'm not sure whether I should support all of them or not.

Could you tell me if you want to keep the possibility for people to provide these via env variables?

* ``pynsee_file_list``
* ``pynsee_use_sdmx``
* ``pynsee_idbank_file``
* ``pynsee_print_url`` (this one I would propose to keep only as config var for sure)

tfardet:
@hadrilec can you also confirm that I should remove [these lines](https://github.com/InseeFrLab/pynsee/blob/master/pynsee/macrodata/_dwn_idbank_files.py#L67-L83) (as you said ``pynsee_idbank_loop_url`` should be removed)

hadrilec:
@tfardet, indeed the while loop should be kept but the usage of the variable `pynsee_idbank_loop_url` should be removed

hadrilec:
> @hadrilec just to clarify something: I still plan to allow some configs to be provided via environment variables but in the list that you provided, I'm not sure whether I should support all of them or not.
> 
> Could you tell me if you want to keep the possibility for people to provide these via env variables?
> 
> * `pynsee_file_list`
> * `pynsee_use_sdmx`
> * `pynsee_idbank_file`
> * `pynsee_print_url` (this one I would propose to keep only as config var for sure)

I would say that the user should be able to modify the value of `pynsee_idbank_file` through an env variable, indeed the download of this file could fail in case insee decides to change the url and then this feature would be a backup at the python session initialization. Same for this variable `pynsee_file_list`

pynsee/metadata/get_definition.py
```diff
@@ -54,14 +56,16 @@ def clean_definition(string):
...
+        except Exception:
             val = None
         return val
 
-    for i in trange(len(ids), desc="Getting data"):
-
-        id = ids[i]
-        query = link + "/" + id
+    for elt in tqdm(
+        len(ids),
```
tgrandje, line 64: Shouldn't that be `for elt in tqdm(ids, ... )` instead?

tfardet:
@tgrandje sorry, I did not have time to work on this lately, other stuff came up, probably won't have time to touch it before next week...

I'm not sure what you mean by "requesting merge": do you mean that you would send specific patches for the config or do you want to merge your whole CI caching PR?

If it's the former, it's completely fine; if it's the later, unless you manage to make your PR really small, I'm not sure it's a good idea.
In the 2nd case it's probably better to make it dependent on this PR (also as a draft) and we'll just merge them in order, no?

tfardet:
OK, I think I'm almost there.
I'd just add a ``Session`` object to the config to reuse throughout the code as discussed with @tgrandje and then stop there.

Usually the session is configured as

```python
    session = requests.Session()
    retry = Retry(connect=3, backoff_factor=1, status_forcelist=[502])
    adapter = HTTPAdapter(max_retries=retry)
    session.mount("http://", adapter)
    session.mount("https://", adapter)
```

in most areas, should this be the default and be used everywhere or are there reasons why this should not be done in some parts of the code @hadrilec ?

tfardet:
Regarding the failing test from [test_request_insee_2](https://github.com/InseeFrLab/pynsee/blob/master/tests/utils/test_pynsee_utils.py#L52), I'm not sure how to handle this:

* apparently in the previous version, pynsee would switch tokens on the fly if the environment was modified
* at the moment it does not, it initializes (giving the priority to the environment token if it's present), then stays that way unless explicitely modified

The tests are not designed around that behavior and are currently failing.

Was the previous behavior intended, and should I keep it that way? @hadrilec @tgrandje

tfardet:
Up ;)
@hadrilec @tgrandje

tfardet:
I won't be able to work on this before October 23, in any case, unless someone chimes in before that, I will change the current ``pynsee`` behavior so it does not react on-the-fly to changes of the environment token and change the tests accordingly.

hadrilec:
sorry @tfardet for the delayed answer, I agree with your proposal.

At the beginning the primary objective was to ensure a seamless usage of the token, in other words the user should only worry about getting the credentials and `pynsee` should handle the token-related tasks. I would say that if the usage of the token fails then another one should be requested from api.insee.fr.

tfardet:
OK, I made some additional changes, moving the credentials into a proper config folder and allowing to save the whole config there.

Regarding the tests, I've tried to fix some of the but it turn out that the SDMX tests require the user to be able to pass invalid credentials. I'm not sure why this was desired, but in the new code, passing an incorrect credential leads to [an error](https://github.com/InseeFrLab/pynsee/pull/162/files#diff-65c2421bd1553be348a8d4d45616ce2f7c83a7d4e2a8836525e22129ae9521aaR115). Should I change that or just remove [these tests](https://github.com/InseeFrLab/pynsee/blob/master/tests/utils/test_pynsee_utils.py#L52)?

tfardet:
@tgrandje @hadrilec sorry for the wait, I think the PR is ready for review now!

Here is the final proposal:

1. upon import, the saved config (including credentials) is loaded from disk, it now lives in the user config folder instead of being in the middle of the user directory
2. environment variables are checked on initial import and replace the config if they exist (for the duration of the script)
3. throughout the code, only the config values are used, this means that only calls to `set_config` can change credentials within a script
4. it is not possible to pass bogus tokens to the configuration, one can only reset it to `None` (or any value with a test returning `False`)
5. it is possible to pass a bogus key or secret by passing it alone, e.g. `set_config("insee_key", "plop")` but that will raise a warning
6. `init_conn` uses `set_config` internally and therefore passing `set_config({"insee_key": key, "insee_secret": secret})` is equivalent to calling `init_conn`, with the specific difference that `init_conn` saves the configuration to disk, making it permanent

pynsee/geodata/_get_geodata.py
```diff
@@ -168,21 +168,38 @@ def _get_geodata(
...
+                    initializer=func_settings, initargs=(args,), processes=Nprocesses
+                ) as pool:                    
+                    list_output = list(
+                        tqdm.tqdm(
+                            pool.imap(func, irange),
+                            total=length,
+                            disable=pynsee.get_config("hide_progress")
+                        ))
+                    
+            except:
```
tfardet, line 188: Avoid bare except:

```suggestion
            except Exception:
```

pynsee/localdata/_find_latest_local_dataset.py
```diff
@@ -37,33 +40,35 @@ def _find_latest_local_dataset(dataset_version, variables, nivgeo, codegeo, upda
...
-                
+
                 if type(df) == pd.core.frame.DataFrame:
                     if len(df.index) == 1:
                         if df['OBS_VALUE'][0] is None:
                             raise ValueError('No data found')
-                        
-            except:            
+
+            except:
```
tfardet, line 63:
```suggestion
            except Exception:
```

pynsee/localdata/_find_latest_local_dataset.py
```diff
@@ -1,33 +1,36 @@
...
 
+
 def _find_latest_local_dataset(dataset_version, variables, nivgeo, codegeo, update):
-    
+
     filename = _hash("".join([dataset_version] + ['_find_latest_local_dataset']))
     insee_folder = _create_insee_folder()
-    file_localdata = insee_folder + "/" + filename 
-    
+    file_localdata = insee_folder + "/" + filename
```
tfardet, line 25:
```suggestion
    file_localdata = os.path.join(insee_folder, filename)
```

pynsee/macrodata/_download_idbank_list.py
```diff
@@ -1,75 +1,70 @@
...
+from pynsee.utils._create_insee_folder import _create_insee_folder
+
+
+def _download_idbank_list(update=False):
+
+    todays_date = date.today()
+    file_to_dwn_default = "idbank_" + str(todays_date.year) + str(todays_date.month)
+
+    insee_folder = _create_insee_folder()
+    file = insee_folder + "/" + _hash(file_to_dwn_default)
```
tfardet, line 21:
```suggestion
    file = os.path.join(insee_folder, _hash(file_to_dwn_default))
```

pynsee/metadata/get_legal_entity.py
```diff
@@ -1,100 +1,102 @@
...
+        codes (list): list of legal entities code of 2 or 4 characters
+
+    Examples:
+        >>> from pynsee.metadata import get_legal_entity
+        >>> legal_entity = get_legal_entity(codes = ['5599', '83'])
+    """
+
+    filename = _hash("get_legal_entity" + "".join(codes))
+    insee_folder = _create_insee_folder()
+    file_legal_entity = insee_folder + "/" + filename
```
tfardet, line 41:
```suggestion
    file_legal_entity = os.path.join(insee_folder, filename)
```

pynsee/utils/_clean_insee_folder.py
```diff
@@ -1,25 +1,25 @@
...
+# Copyright : INSEE, 2021
+
+import platformdirs
+import os
+
+
+def _clean_insee_folder():
+
+    local_appdata_folder = platformdirs.user_cache_dir()
+    insee_folder = local_appdata_folder + "/pynsee" + "/pynsee"
```
tfardet, line 11: There was pynsee twice, was that normal?

```suggestion
    insee_folder = os.path.join(local_appdata_folder, "pynsee")
```

pynsee/utils/_clean_insee_folder.py
```diff
@@ -1,25 +1,25 @@
...
+        # exclude directories
+        list_file_insee = [
+            file
+            for file in list_file_insee
+            if os.path.isfile(os.path.join(insee_folder, file))
+        ]
+
+        if len(list_file_insee) > 0:
+            for f in list_file_insee:
+                os.remove(insee_folder + "/" + f)
```
tfardet, line 25:
```suggestion
                os.remove(os.path.join(insee_folder, f))
```

pynsee/utils/_create_insee_folder.py
```diff
@@ -1,47 +1,45 @@
...
+            os.mkdir(insee_folder)
+
+        # create internal folder
+        # if folder is not None:
+        #     insee_folder = insee_folder + '/' + folder
+        #     if not os.path.exists(insee_folder):
+        #         os.mkdir(insee_folder)
+
+        # test if saving a file is possible
+        test_file = insee_folder + "/" + _hash("test_file")
```
tfardet, line 36:
```suggestion
        test_file = os.path.join(insee_folder, _hash("test_file"))
```

tfardet (commented):
@hadrilec I'm not sure what happened to make some files appear completely overwritten... it makes the changes quite hard to find.
Overall I'm a bit worried this PR is becoming almost too large to handle, so if there are things that you added that can be done in a separate PR, that might help.

If not, well, let's hope that @tgrandje can give us a new pair of eyes on this to make sure we did not make any big mistake :grin: 

I made a few proposal for minor changes (except for one that looks like a bug, but maybe I'm mistaken)

hadrilec:
yes, I agree, I should have made the changes elsewhere, let's see if I time later on to decouple the changes

I deleted the function `_wait_api_query_limit`, because we drew the [conclusion](https://github.com/InseeFrLab/pynsee/issues/164#issuecomment-1699508835) that is was not needed anymore, and after having performed several tests it is indeed the case.

regarding, the issue on "metadonnees" I understand that the connection is lost at some point, I an going to add a by default user agent to the header of the requests to see if it solves the problem as it is the advice described here: 
https://stackoverflow.com/questions/61827856/python-connectionerror-httpsconnectionpoolhost-api-foursquare-com-port-443

pynsee/utils/_request_insee.py
```diff
@@ -71,10 +71,16 @@ def _request_insee(
     if api_url is not None:
         token = pynsee.get_config("insee_token")
 
+        try:
```
tfardet, line 74: replace by ``os.environ.get("USERNAME", "username")``

pynsee/utils/config.py
```diff
@@ -152,9 +152,15 @@ def _register_token(
             "!!! Token is missing, please check that insee_key and "
             "insee_secret are correct !!!")
     else:
+        try:
```
tfardet, line 155: see above

pynsee/utils/config.py
```diff
@@ -3,11 +3,13 @@
 import json
 import logging
 import os
-import requests
+#import requests
 import urllib3
 
 from typing import Any, Optional, Union
 
+from pynsee.utils._request_insee import _request_insee
```
tfardet, line 11: You cannot do that, _request_insee depends on config files.
You could import it within the function if it is really necessary, but I'm not sure why it would be, could you explain?

hadrilec:
Hi @tfardet,

I executed the code below and whenever `pynsee` is imported queries are triggered to check whether the credentials are working well. I have the feeling that this snippet is also used by many other functions, hence the number of sessions and requests created might increase a lot without fulfilling a user need. I believe this explains why the tests are failing.

Could you please make sure the function `_register_token` is used only by `init_conn` and is not used by any other function?

`
import logging
import sys
logging.basicConfig(stream=sys.stdout,
                    level=logging.DEBUG,
                    format="%(message)s")

from pynsee import *
`

tfardet:
@hadrilec ah, sorry, maybe I missed something, will check it out!

tfardet:
@hadrilec it runs just fine locally,  and the code seems OK... can you post what seems wrong on your side?

hadrilec:
well, I have the same but I think this should not be the targeted behaviour. From a user point of view, I think it is better to start testing the connection with the api through the init_conn function and not with the import package. In addition, I suspect that as many functions seem to rely on set_config, the function testing the connection to all apis might be triggered all over again and again, is that right or is there something I misunderstand?

tfardet:
1. Regarding the current behavior:
    - I think that the whole point of providing environment variables is to *not* have to call ``init_conn`` so we do have to check and initialize the token from them on import
    - if the env variables are not set and you ran ``init_conn`` in the past, you can check that no requests are made
2. Regarding ``get/set_config``:
    - you can check that ``get_config`` never makes requests
    - ``set_config`` only calls ``_register_token`` if a valid-looking "token" entry or a "key/secret" pair is passed

tfardet:
That being said, tests are now taking 4h to run, so there's definitely something wrong, I just don't see what it is, I'll investigate

tfardet:
@hadrilec honestly, this PR has become unmanageable IMO ^^"
Could you please redo your fix for the tests and move to 3.8/3.11 in another PR?

Once we merge that (hopefully the tests will be back to less than 1h), I'll redo a PR from that one.
Otherwise it's going to take forever to debug this.

hadrilec:
Thanks a lot @tfardet for the second PR! 
Indeed, you are right, I guess it is the backup solution in case the multiprocessing fails that takes forever to run.

hadrilec:
@tfardet @tgrandje, guys let me know if I can do anything to move forward

tfardet:
@hadrilec  given the amount of changes to the code since I started, I need to start this again from scratch.
Unfortunately, I have too many things on my plate at the moment, so I won't be able to work on this before April at best, and more probably May.

PR #166, opened 2023-08-08
Support range patterns of SIRENE API

### LINKED ISSUE #165: search_sirene does not support range patterns
Because of the automatic split in [search_sirene](https://github.com/InseeFrLab/pynsee/blob/master/pynsee/sirene/search_sirene.py#L180), the function cannot handle [range requests](https://www.sirene.fr/static-resources/htm/multi_plages.html).

E.g.

```python
pynsee.search_sirene(variable = ["codeCommuneEtablissement", "trancheEffectifsEtablissement"],
                     pattern = ['77083', "[12 TO 53]"], kind = 'siret')
```

Will propose a fix soon.

tfardet:
Fixes #165 by supporting the ``[X TO Y]`` and ``{X TO Y}`` patterns (cf. [doc](https://www.sirene.fr/static-resources/htm/multi_plages.html)).

NB: I used ``\w`` for the pattern matching, not sure whether we need to support special characters, let me know if it should be something else, or even ``.``.

### DIFF
--- a/pynsee/sirene/search_sirene.py
+++ b/pynsee/sirene/search_sirene.py
@@ -177,8 +177,18 @@ def search_sirene(
         else:
             phntc_string = ".phonetisation"
 
-        # if pattern has several words, split and put mutiple conditions with OR
-        patt = re.sub(r"\s+", "|", patt)
+        # if pattern has several words, split and put mutiple conditions
+        # using OR, unless it's a range condition, cf.
+        # https://www.sirene.fr/static-resources/htm/sommaire.html
+        if not (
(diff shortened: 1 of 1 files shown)


hadrilec:
is this PR finalised, can I merge it?

tfardet:
@hadrilec there's just the question of special characters: if we need to support anything other than a-z, A-Z, 0-9 and _, then I need to make some changes; otherwise we're good to go.
This requires more in-depth knowledge about the content of the SIRENE database than what I possess, hence my question.

PR #171, opened 2023-08-22
Update ``get_location`` documentation and add caching

### LINKED ISSUE #168: Add an `exact_location` field for SIRENE DataFrames with `get_location`
Right now, there is no obvious way to check whether the location returned by [``get_location``](https://github.com/InseeFrLab/pynsee/blob/master/pynsee/sirene/get_location.py#L40) is exact or not.

I'd propose to add a column with a title like "exact_location" that would be a boolean with ``True`` if the [first call to `get_location_openstreetmap`](https://github.com/InseeFrLab/pynsee/blob/master/pynsee/sirene/get_location.py#L140) succeeds and ``False`` otherwise.

tfardet:
Fixes #168 and adds caching to [``_get_location_openstreetmap``](https://github.com/InseeFrLab/pynsee/blob/master/pynsee/sirene/_get_location_openstreetmap.py).

I also cleaned up some imports and bare ``except`` clauses.
Finally, I moved the ``get_location`` directly as a method and removed its former file.

You can ignore the two test files, I just checked them and my editor autoremoves trailing whitespaces.


tfardet:
Failure in 3.10 seems unrelated to the PR (call to [get_geodata](https://github.com/InseeFrLab/pynsee/pull/171/files#diff-3361e550626583456be86e17519be40a246b1d8b8274c89dd8dc9963cc67bd5dL65))  which has not been modified... not sure whether it's worth restarting the test... @hadrilec I'll let you decide on that.

PR #173, opened 2023-09-02
bugfix - import object instead of module

hadrilec:


tfardet:
Ah, yeah, I'm the one who introduced this bug... there is still two imports missing in ``test_pynsee_geodata/sirene.py``.
Sorry about that! Don't know why it did not appear in the tests...
I did not expect the module to take precedence over the object when they have the same name, this really makes it a pain to manage imports...

tfardet:
Wait, @hadrilec I just noticed that CI is [checking master](https://github.com/InseeFrLab/pynsee/actions/runs/6061118101/job/16445950584?pr=173#step:2:384) instead of the latest PR commit

Issue #175, opened 2023-11-06
Population Map example


elGringo11:
Hello   have tried to run the example Population Map and  got a bunch of errors:
PS C:\Users\XXX\Desktop\INSEE_API> & "C:/Program Files/Python311/python.exe" c:/Users/XXX/Desktop/INSEE_API/carte.py
API query number limit reached - function might be slowed down
Thanks for your help.

hadrilec:
hello, thanks for your feedback, could you please provide a reproducible example in this issue?

elGringo11:
Hi,

Is that you expect?

Thanks


```
from pynsee.utils.init_conn import init_conn
init_conn(insee_key="Tmr9TcN1WU2Y7xjjK4cCfrGolCka", insee_secret="viZDkyMqdCwbGw647xL2NYVJAr4a")


from pynsee.geodata import get_geodata_list, get_geodata, GeoFrDataFrame


import math
import geopandas as gpd
import pandas as pd
from pandas.api.types import CategoricalDtype
import matplotlib.cm as cm
import matplotlib.pyplot as plt
import descartes

import warnings
from shapely.errors import ShapelyDeprecationWarning
warnings.filterwarnings("ignore", category=ShapelyDeprecationWarning)

import logging
import sys

logger = logging.getLogger()
logger.setLevel(logging.INFO)
formatter = logging.Formatter('[%(filename)s:%(lineno)s - %(funcName)20s() ] %(message)s')

file_handler = logging.FileHandler('mylogs.log')
file_handler.setLevel(logging.DEBUG)
file_handler.setFormatter(formatter)

logger.addHandler(file_handler)


# get geographical data list
geodata_list = get_geodata_list()
# get departments geographical limits
com = get_geodata('ADMINEXPRESS-COG-CARTO.LATEST:commune')

mapcom = gpd.GeoDataFrame(com).set_crs("EPSG:3857")

mapcom = mapcom.to_crs(epsg=3035)
mapcom["area"] = mapcom['geometry'].area / 10**6
mapcom = mapcom.to_crs(epsg=3857)

mapcom['REF_AREA'] = 'D' + mapcom['insee_dep']
mapcom['density'] = mapcom['population'] / mapcom['area']

mapcom = GeoFrDataFrame(mapcom)
mapcom = mapcom.translate(departement = ['971', '972', '974', '973', '976'],
                          factor = [1.5, 1.5, 1.5, 0.35, 1.5])

[...]

PR #178, opened 2023-11-19
Test Python 3.8/3.11 + backup multiprocessing to get geodata

### LINKED ISSUE #172: Test against relevant python versions
I'd propose to stop testing against python 3.7, as it already reached [end of life](https://endoflife.date/python) and start using 3.11 instead.

I also wonder whether it is necessary to test against all intermediate versions and whether oldest and newest alone would not be sufficient (especially given how long the tests take to run). No strong opinion, though, just wondering whether it's worth it.

tfardet:
Fixes #172 and #175 

@hadrilec are these the main changes you had included in #162? If not, feel free to include whatever is missing.


tfardet:
@hadrilec Well, it seems we now know where the 4h-long tests come from ^^"

tfardet:
@hadrilec I won't have time to look at this for the moment, so if you can have a look, it would be worth checking what makes the multiprocessing fail.
I also wonder whether this is the best way to handle this and if we shouldn't rather move to something like [this](https://stackoverflow.com/a/57129241/5962321), might be just as fast (faster?) and more reliable.

hadrilec:
Thanks for the link, I will have a look. I agree async should be the way to go.

hadrilec:
I did a mistake by pushing commits from the other PR, my bad, I will clean this up

tfardet:
@hadrilec do you know when you'd have time to finish this part?

PR #192, opened 2024-04-03
Move to parquet and platformdirs

tfardet:
Use parquet instead of pickle to store datasets (around 100 times faster), should be supported since pandas 0.21 and we require 0.24.

Move to platformdirs for cache folder, etc, as appdirs is unmaintained.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -16,7 +16,7 @@ sphinxcontrib-svg2pdfconverter
 pandas>=0.24.2
 tqdm>=4.56.0
 requests>=2.25.1
-appdirs>=1.4.4
+platformdirs
 unidecode>=1.2.0
 datetime>=3.5.9
 pathlib2>=2.3.5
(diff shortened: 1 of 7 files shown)


tfardet:
Apparently I need to add pyarrow as a dependency for parquet to run (I thought it was installed by default with pandas now, but apparently not), let's see if it runs fine this time.
@hadrilec  I hope the additional dependency is OK with you, I really think parquet is the way forward in terms of storing data (both to save space and fast IO).

tfardet:
@hadrilec you can have a look, tests are passing now that SIRENE is fixed

tfardet:
@hadrilec let me know if you need further info to validate that PR

hadrilec:
@tfardet thanks a lot, do you think it would be possible to move pyarrow to the optional packages? meaning to the extras_require list in setup.py file. In doing so, we would allow the users to use pyarrow if they want and it would not add a "hard" dependency on pyarrow. if this too much trouble, let me know

tfardet:
> do you think it would be possible to move pyarrow to the optional packages?

This is theoretically possible but I would strongly advise against it: this would require two different codes to handle file for people with and without pyarrow and making pyarrow non-default would mean that most people would not get the benefits from the new parquet format...

Is there a specific issue with pyarrow that you are aware of?
As far as I know they provide wheels so installation should never be an issue...

PR #193, opened 2024-04-04
Fix sirene endpoint url

elishowk:
I think the endpoint `/sirene3/V3.11/` is retired. In fact, INSEE's api documentation does not mentions it. May be I'm not aware of the goal of the `/sirene3/V3.11/`

# Reproduction

## Works

`http https://api.insee.fr/entreprises/sirene/V3/siret\?q\=\(dateDernierTraitementEtablissement:\[2024-04-03T11:59%20TO%202024-04-04T11:59\]\)\&nombre\=1\&curseur\=\*`

## Does not work (503 error)

`https://api.insee.fr/entreprises/sirene3/V3.11/siret\?q\=\(dateDernierTraitementEtablissement:\[2024-04-03T11:59%20TO%202024-04-04T11:59\]\)\&nombre\=1\&curseur\=\*`

### DIFF
--- a/pynsee/sirene/_request_sirene.py
+++ b/pynsee/sirene/_request_sirene.py
@@ -29,7 +29,7 @@ def _request_sirene(query, kind, number=1001):
     else:
         raise ValueError("!!! kind should be among : siren siret !!!")
 
-    INSEE_api_sirene_siren = "https://api.insee.fr/entreprises/sirene3/V3.11"
+    INSEE_api_sirene_siren = "https://api.insee.fr/entreprises/sirene/V3.11"
     number_query_limit = 1000
 
     number_query = min(number_query_limit, number)


tfardet (approved):

hadrilec:
V3.11 is mentioned here: 
https://api.insee.fr/catalogue/site/themes/wso2/subthemes/insee/pages/item-info.jag?name=Sirene&version=V3.11&provider=insee
I need to perform further checks to understand why the package has some issues, I have been told by INSEE that the data was not complete that is why during the testing phase the tests were failing, my assumption was that after the go-live in production the tests would work as the data would be complete. Apparently, there are still issues, so I will try to have a look in the coming days.

tfardet:
@hadrilec not sure what your previous message meant: do you need to check with people at INSEE what happened or is this the right fix and we can just merge it?

hadrilec:
as @elishowk mentioned, indeed, I agree after a second look I dont find it, then I might have left a typo. Thanks a lot for the fix

PR #199, opened 2024-06-02
Save decorator used in download module + bug fix in get_dataset

hadrilec:


pynsee/download/_download_store_file.py
```diff
@@ -49,18 +45,11 @@ def _download_store_file(id: str, update: bool):
                 potential values are: {suggestions}"
 
         raise ValueError(error_message)
+        
+    filename = tempfile.mkdtemp() + "tempfile"
```
tfardet, line 49: use ``os.path.join`` to make sure this will always work on all systems

pynsee/download/_load_data_from_schema.py
```diff
@@ -136,4 +136,17 @@ def _load_data_from_schema(
...
+    try:
+        if telechargementFichier["result"]["zip"] is True:
+            list_files += [dataPathFile,
+                           f"{zipDirectory}/" + telechargementFichier["file_archive"],
+                           telechargementFichier["file_archive"]]
+        
+        for f in list_files:
+            if os.path.exists(f):
+                os.remove(f)
+    except:
```
tfardet, line 149: Better not to have bare excepts: if I'm not mistaken, the only potential error here is from ``os.remove``, right?
In that case it would be better to wrap that one line, so that if one file fails, later files are still removed.
The error raised should be ``FileNotFoundError`` if the file does not exist, there may be other if the user does not have write access.

tfardet (changes requested):
Thanks for the fix, just a few minor comments

pynsee/download/_load_data_from_schema.py
hadrilec: indeed, I was mostly worried about the access rights

hadrilec: this commit d201b0795428266de9ac0b95e4f7b6179400f2a7 addresses the issue

pynsee/download/_download_store_file.py
hadrilec: this commit d201b0795428266de9ac0b95e4f7b6179400f2a7 addresses the issue

tfardet:
@hadrilec the tests look a bit weird, did you cancel the 3rd one manually?

hadrilec:
checks were performed successfully, there are still warnings in 3.11 tests that should be addressed

tfardet (approved):
OK, I don't know what happened in the middle, but as long as it's solved, all is well.

PR #210, opened 2024-08-11
Use chmod 600 config file in user config dir

tfardet:
Use a 600-mode config file for better security of the credentials and put it in a proper config dir.


tfardet:
@hadrilec I had to change the SIRENE URL in init_conn; since the default API is now 3 and the old V3 link did not work anymore. I also removed the 3.11 from SIRENE requests to avoid issues if there is a new point release.
Let me know if you want me to make a separate PR in case you don't want to merge the new credentials management right away.

hadrilec:
hi, thanks for spotting this bug, a second PR for only the bug would be great please

PR #214, opened 2024-08-31
Bugfix - use http proxy from env variables

tgrandje:
Small PR to :
* handle corporate proxies when pynsee connection is managed through os environment variables;
* use proxies during tests ran on a local machine;
* restore both proxies and credentials after tests are run on a local machine (if keys where stored on disk).

### DIFF
--- a/pynsee/utils/_get_credentials.py
+++ b/pynsee/utils/_get_credentials.py
@@ -15,7 +15,7 @@ logger = logging.getLogger(__name__)
 
 
 def _get_credentials() -> Dict[str, str]:
-    '''
+    """
     Try to load credentials.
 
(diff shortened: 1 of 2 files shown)


tfardet (approved):
looks good to me; as long as the tests pass, I think we can merge this

tgrandje:
@tfardet obviously the tests fail (again) and this is not due to this PR...
Just to be sure, I'll try to run those on my professionnal laptop monday (behind corporate proxy).

tgrandje:
This passes on my professional laptop (python 3.10), including the `utils` tests (I did not bother pushing a commit to force testing on all python version, but I ran those) .

PR #215, opened 2024-08-31
Fix mismatched exception types

tgrandje:
There was a mismatch of exception types in `_request_insee.py` (both RequestException and ValueError).

This PR uniformizes all exception and update tests accordingly (I encountered one test not appropriately set while working on the PR for tests' caching).

### DIFF
--- a/pynsee/localdata/get_geo_list.py
+++ b/pynsee/localdata/get_geo_list.py
@@ -1,17 +1,16 @@
 # -*- coding: utf-8 -*-
 # Copyright : INSEE, 2021
 
-import os
 import pandas as pd
 from tqdm import trange
 import numpy as np
-from requests.exceptions import RequestException
 
 from pynsee.utils._paste import _paste
 from pynsee.localdata._get_geo_relation import _get_geo_relation
 from pynsee.localdata._get_geo_list_simple import _get_geo_list_simple
 
 from pynsee.utils.save_df import save_df
+
 
 @save_df(day_lapse_max=90)
 def get_geo_list(geo=None, date=None, update=False, silent=False):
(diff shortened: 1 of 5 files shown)


pynsee/localdata/get_geo_list.py
```diff
@@ -94,7 +93,7 @@ def get_geo_list(geo=None, date=None, update=False, silent=False):
                     date=date,
                     type=type_geo,
                 )
-            except RequestException:
+            except ValueError:
```
tfardet, line 96: Could you specify why/where the value error comes from in the ``_get_geo_relation``? (maybe adding a comment there?)
I suspect you ran into the error and tracked it down but it's really not obvious from reading the function, so having it explicitely written in the docstring would be nice.

tgrandje: After some thought, I'm not sure that I understand that comment 🙁 .

The primary object of this PR was to ensure unity of exceptions' types generated by `_request_insee` (whether we're using a token or not, whether we use api_url or sdmx_url). As `get_geo_list` runs `_get_geo_relation` which runs `_request_insee`, the exception's type should be the same.

I've just added a simple comment and a docstring to _get_geo_relation (which was missing indeed) , was this what you had in mind?

tfardet: that's fine, this indeed answers my question

tfardet:
I restarted the tests because I think they should not have failed

tgrandje:
> I restarted the tests because I think they should not have failed

Yes I think this might be due to too much workflows at the same time (given that they share the same API keys credentials : until we've patched requests, that's gonna stay unreliable...)

tfardet (approved):
I think this is simple enough to merge even though the tests are failing.
@hadrilec is that OK with you?

Issue #203, opened 2024-08-06
Multiprocessing sometimes freezes get_geodata


tgrandje:
This is something I have already encountered on a linux machine.
It seems to be linked to the multiprocessing freeze support not being properly handled, but the user might not even know of it without heavy debugging.

It occured to me again while debugging [this issue](https://github.com/InseeFrLab/pynsee/issues/202) and this is definitely not easy peasy...

As multiprocessing does not seem to be of any value here (compared to multithreading), I propose to switch to pebble ThreadPool which has a close enough API.

tfardet:
Why go for pebble rather than multithreading?

tgrandje:
To be honest? Mostly lazyness... 😊

And to be more thorough: I like pebble, it has a unified API for both multiprocessing and multithreading which makes switching from one to the other really simple. I've had a go on concurrent.futures in the past, and I don't have very good memories of it (it is not straightforward as you need to submit the function for each args, then mark the results as completed...).

Do you object to pebble? (Btw: I've never encountered a single trouble installing it.)

tfardet:
> Do you object to pebble

I don't know anything about pebble, I just wanted to know whether there was a strong reason to add new dependencies.

> I just saw the two hanging PR. Are they not linked to that same problem?

Honestly I don't remember, but there was definitely an issue with that code so rewriting it is definitely a good idea!

hadrilec:
I understand this issue is outdated, shall we close it?

tfardet:
yes

PR #216, opened 2024-08-31
Enhancement/switch multiprocessing to multithreading

tgrandje:
This is a suggestion to switch from multiprocessing to multithreading for the geodata package:
* there are sometimes troubles using multiprocessing with windows machines, which are not easily debugged;
* as far as I have seen, there is not much interest for multiprocessing as no heavy computation is done, and (I think) we're perfectly safe with multithreading.

This needed to rewrite the creation of the session object, which is now created outside the pool scope (instead of global objects created at threads/processes creation).


tgrandje:
This is derived from [this PR](https://github.com/InseeFrLab/pynsee/pull/204). As the geoparquet issue is still waiting for concensus, I chose to open a new PR working on this particular subject instead.

As already said [here](https://github.com/InseeFrLab/pynsee/issues/203#issuecomment-2272863818), the choice of using pebble is mostly lazyness. The advantage is that it's API si very similar to that of multiprocessing (instead of something like concurrent.futures for instead). I've never had any trouble installing pebble, so I think it is reliable enough for this...

tgrandje:
Just moved from pebble to concurrent.futures, capitalizing on the latest PR from @tfardet .

tgrandje:
Note: this PR should only be examined after merging https://github.com/InseeFrLab/pynsee/pull/246

hadrilec:
hi most machines do not have many cores, that is why as a rule of thumb based on my personal computer capabilities I chose 6 :D also, if you work on a way bigger machine usually you share resources so the objective was not to take all resources from others
of course if we design the data download in a different way and only 7 queries are needed, then multithreading is the way to go and whatever is related to multiprocessing can be dropped

tfardet (approved):
Nice, thanks!

PR #217, opened 2024-09-07
new catalog module using Melodi API

hadrilec:


tgrandje:
@hadrilec I'm closing this and starting anew (I think this one was started before we updated the new session object, so a lot has changed). New PR is [here](https://github.com/InseeFrLab/pynsee/pull/306)

PR #224, opened 2024-10-14
Uniformize silent kwargs

tgrandje:
* add silent kwarg to all function using the save_df decorator
* uniformize the default silent value everywhere
* inherit silent values from parent functions everywhere


hadrilec:
tbh, I am not sure that I want everything to be silenced by default

tgrandje:
@hadrilec  well, I was also not in favor of it, but most queries were already silenced. If we're ok that the default behaviour should be the invert, I'll do it. I still think we should have a consistent API, but that's all.

tgrandje:
Just merged inspect-github-timeouts into this PR. We still need to decide whether the default should be silent=True or silent=False for a consistent API. (Either way is ok with me 😉 )

tgrandje:
_Note: I cancelled the workflow to avoid having 2 (or 3) at the same time, but we'll need to restart those when possible._

tgrandje:
Just added some improvements. To summarize it all:
* small bugfix in requests_session in case of missing variable from environ (I can move it to a separate PR if you prefer)
* added some docstrings to `save_df` and `_save_dataframe`
* also removed the (now) completly unused `parquet` argument from both functions
* add some missing "silent" kwargs (argument now available for each function using the `save_df` decorator)
* fix some cascading values for `silent` in child functions
* allow to trigger the `silent` config globally using `PYNSEE_SILENT_MODE` environ constant. Not sure were we should document this though... For now, it's in the docstrings.

Note: this PR should only be examined after merging https://github.com/InseeFrLab/pynsee/pull/246

tfardet (approved):

Issue #231, opened 2025-01-17
Bug in macrodata/_dwn_idbank_files


tgrandje:
There is something strange going on on [those lines](https://github.com/InseeFrLab/pynsee/blob/9dc111f80be32451a15ad88e0826eb5c1fe0414b/pynsee/macrodata/_dwn_idbank_files.py#L38-L48).

1. should it be retrieved on [line 44](https://github.com/InseeFrLab/pynsee/blob/9dc111f80be32451a15ad88e0826eb5c1fe0414b/pynsee/macrodata/_dwn_idbank_files.py#L44), `data` is never used (as `files` is never empty) ;
2. anyway, `file_to_dwn` set as constant [here](https://github.com/InseeFrLab/pynsee/blob/9dc111f80be32451a15ad88e0826eb5c1fe0414b/pynsee/macrodata/_dwn_idbank_files.py#L41) is not available as of today
3. further more, `_dwn_idbank_file` needs a `session` argument, so this is always trigerring the exception.

@hadrilec can you detail what the expected behaviour should be ?

I'm not really sure it this sample should be dropped altogether or if there is a reason still to keep it...

hadrilec:
hi, the intention was to allow the user the ability to change manually the file retrieved from insee.fr. This list is updated manually on insee side, so it is not unlikely that the file change name. These lines should then considered as a backup before a permanent fix is released in a new version of the package. The code first try to use the file the user intentionally set and then use dates to create potential correct urls. The quality of this snippet of code could of course be better. Shall I provide you with more details?

tgrandje:
OK, so the missing part is the `session` argument to `_dwn_idbank_file`. I'll fix that in the PR fixing the github actions timeouts (makes sense as this is already messing with the requests sessions).

tgrandje:
done

PR #232, opened 2025-01-17
Resolve GitHub timeouts

tgrandje:
* create a class to manage every http requests (wether to the INSEE APIs, the IGN's or anything else) ;
* set a default timeout for each query (might be needed to extend this timeout later, but this seems ok in regard to existing tests)
* run all test on any python version
* run tests only on py 3.9 & 3.12
* add a __version__ for pynsee, for 2 reasons :
    * traceability (giving the extension of current changes, I suggest the 0.2.0 version for pynsee)
    * and the IGN's geoplatform doesn't allow for a User-Agent without a "/blah" part, which is usually reserved to the current application's version.
* alter utils tests
* remove 404 tracking from INSEE's API results as this is a "no result found" code, not an error

I might have blundered here or there on the display of log entries, but I hope not too much.
This should be checked thoroughly and the black formatting will not make this an easy task, sorry about that...


tgrandje:
Just added resolution to https://github.com/InseeFrLab/pynsee/issues/231

tgrandje:
Just in case we agree on that, I've also opened [another branch](https://github.com/InseeFrLab/pynsee/tree/session-with-rate-limiters) which allows to track each API's current rate separatly using requests_ratelimiter.

tgrandje:
Also :

* take care of https://github.com/InseeFrLab/pynsee/issues/238 
* fix some paths, still displaying version number in URLs, which are not displayed in the current portal (anymore ?)
* prune some unnecessary code while testing the API's correct access and make use of the new session object's built-in exception

tgrandje:
Not a 100% sure of the latest commit.
I sure resolved some bugs, but I'm currently not working with a static proxy and cannot be certain of the proper proxy behaviour (I'll try to reinstall the environment on another machine). Please fill free to advice on it!

tgrandje:
@tfardet that's as good as it's gonna be until you can check this.

The diff format is still gonna be a pain to look at as :
* there are still parts where using the "with" statements fooled much of git, 
* as well as parts where I honestly don't know why the whole file is marked as changed (might be me switching between WSL and windows on two different machines?).

I also added the cache limiters I speak of yesterday.

pynsee/utils/init_conn.py
```diff
@@ -24,12 +20,12 @@ def opener(path, flags):
 
 
 def init_conn(
-    sirene_key: str, http_proxy: str = "", https_proxy: str = ""
+    sirene_key: str = None, http_proxy: str = None, https_proxy: str = None
```
tfardet, line 23: these needs to be ``str | None``, then, but it's not consistent with your test for sirene_key == "" above, we should pick one or the other

pynsee/utils/_get_credentials.py
```diff
@@ -18,21 +18,25 @@ def _get_credentials() -> Dict[str, str]:
...
     """
     key_dict: Dict[str, str] = {}
 
-    sirene_key = os.environ.get("sirene_key")
+    def get_env_case_insensitive(x: str):
+        "Fetch an environment variable (case insentitive)"
+        val = os.environ.get(x)
+        if val is None:
+            return os.environ.get(x.upper())
+        return val
```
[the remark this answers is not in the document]
tfardet: or ``return os.environ.get(x, os.environ.get(x.upper(), ""))`` if you want a string whatever happens

pynsee/utils/init_conn.py
```diff
@@ -55,74 +51,69 @@ def init_conn(
...
+
+            try:
+                if api == "Sirene" and not session.sirene_key:
+                    # the user is probably not expecting to use SIRENE anyway:
+                    # simple warning and jump to next API in order to avoid
+                    # urllib retries
+                    logger.warning(
+                        f"Remember to subscribe to {api} API on api.insee.fr "
+                        "if you ever want to use it (type `help(init_conn)` "
+                        "to know more about this)"
```
tfardet, line 76: No need for an f-string, this is SIRENE only

pynsee/utils/init_conn.py
```diff
@@ -55,74 +51,69 @@ def init_conn(
...
+                if exc.response.status_code == 404:
+                    raise RuntimeError(
+                        f"Could not reach {api} at {api_url}, please get in "
+                        "touch if the issue persists."
+                    )
+                elif results.status_code != 200:
+                    logger.critical(
+                        f"Please subscribe to {api} API on api.insee.fr !\n"
+                        f"Received error {exc.response.status_code}: "
+                    )
```
tfardet, line 98: so should this only happen with SIRENE? I'm not sure this is properly handled at the moment in terms of what we tell the person who will see the message...

pynsee/utils/init_conn.py
```diff
@@ -134,8 +125,8 @@ def init_conn(
     # save config
     config = {
         "sirene_key": sirene_key,
-        "http_proxy": http_proxy,
-        "https_proxy": https_proxy,
+        "http_proxy": http_proxy if not http_proxy == "" else "DIRECT",
+        "https_proxy": https_proxy if not https_proxy == "" else "DIRECT",
```
tfardet, line 129: What's that?
Usually the default value is "" or None, what is the meaning of "DIRECT"?
Also the preferred syntax would be ``http_proxy or "DIRECT"``

pynsee/utils/requests_session.py
```diff
@@ -0,0 +1,360 @@
...
+        # 429: "Too Many Requests : allocated quota overloaded",
+        500: "Internal Server Error ",
+        503: "Service Unavailable",
+    }
+
+    def __init__(
+        self,
+        http_proxy: str = None,
+        https_proxy: str = None,
+        sirene_key: str = None,
```
tfardet, line 46: ``str | None``

pynsee/utils/requests_session.py
```diff
@@ -0,0 +1,360 @@
...
+
+        if any(x is not None for x in (http_proxy, https_proxy, sirene_key)):
+            # set credentials for a new pynsee connection
+            credentials = {
+                "http_proxy": http_proxy,
+                "https_proxy": https_proxy,
+                "sirene_key": sirene_key,
+            }
+        else:
+            credentials = _get_credentials()
```
tfardet, line 90:
```suggestion
        credentials = _get_credentials()
        
        if http_proxy is not None:
            credentials["http_proxy"] = http_proxy
        
        if https_proxy is not None:
            credentials["https_proxy"] = https_proxy
        
        if sirene_key is not None:
            credentials["sirene_key"] = sirene_key
```

so that the default values are used (otherwise setting proxy will drop the SIRENE key

pynsee/utils/requests_session.py
```diff
@@ -0,0 +1,360 @@
...
+        self.proxies.update(proxies)
+
+        username = os.environ.get("USERNAME", "username")
+        version = pynsee.__version__
+        useragent = {"User-Agent": f"python_pynsee_{username}/{version}"}
+
+        # Note : geoplatform seems to impose the "/version" to the user-agent
+        self.headers.update(useragent)
+
+        self.sirene_key = credentials.get("sirene_key", None)
```
tfardet, line 105: None is not necessary, it's the default value for ``get``

pynsee/utils/requests_session.py
```diff
@@ -0,0 +1,360 @@
...
+        self.headers["X-INSEE-Api-Key-Integration"] = self.sirene_key
+
+    def _mount_adapters(self):
+
+        # default retries adapter
+        retry_adapt = Retry(
+            total=7, backoff_factor=1, status_forcelist=[429, 502, 503, 504]
```
tfardet, line 112: could you add the meaning of the missing codes in the list at the beginning of the class?

pynsee/utils/requests_session.py
```diff
@@ -0,0 +1,360 @@
...
+        )
+        adapter = HTTPAdapter(max_retries=retry_adapt)
+        self.mount("https://", adapter)
```
tfardet, line 115:
```suggestion
        self.mount("http://", adapter)
```

pynsee/utils/requests_session.py
```diff
@@ -0,0 +1,360 @@
...
+        self.mount("https://", adapter)
+
+        insee_folder = _create_insee_folder()
+        rate_folder = os.path.join(insee_folder, "rate_limiter")
+        if not os.path.exists(rate_folder):
+            os.makedirs(rate_folder)
```
tfardet, line 121:
```suggestion
        os.makedirs(rate_folder, exist_ok=True)
```

pynsee/utils/requests_session.py
```diff
@@ -0,0 +1,360 @@
...
+
+    def request_insee(
+        self,
+        api_url=None,
+        sdmx_url=None,
+        file_format="application/xml",
+        print_msg=True,
+        raise_if_not_ok=False,
+    ):
+        "Performs a query to insee, either through API or sdmx_url"
```
tfardet, line 210:
```suggestion
        """Performs a query to INSEE, either through API or sdmx_url"""
```

pynsee/utils/requests_session.py
```diff
@@ -0,0 +1,360 @@
...
+                    raise_if_not_ok=raise_if_not_ok,
+                    print_msg=print_msg,
+                )
+
+        except ValueError as e:
+
+            if self._query_is_valid_sirene_call(api_url):
+                # Log the error as a SIRENE API call can not function without
+                # key
+                logger.critical(e)
```
tfardet, line 244: why is this necessary? I don't think ``self._request_api_insee`` raises a ``ValueError`` so this is only for line 227, isn't it? why not directly log the message?

pynsee/utils/requests_session.py
```diff
@@ -0,0 +1,360 @@
...
+
+        except requests.exceptions.RequestException:
+            if sdmx_url:
```
tfardet, line 247: I don't remember how ``sdmx_url`` is supposed to be used but this is never reached if ``api_url`` is not provided, is that the expected behavior?
Basically if called without ``api_url``, the function just returns None.

cc @hadrilec if you have time I think it would be good if you could just check this one file

pynsee/utils/requests_session.py
```diff
@@ -0,0 +1,360 @@
...
+                if print_msg:
+                    logger.critical("SDMX web service used instead of API")
+                result = self._request_sdmx_insee(
+                    sdmx_url, raise_if_not_ok=raise_if_not_ok
+                )
+
+        return result
+
+    def _query_is_valid_sirene_call(self, url):
+        "check if an URL is a valid API call to SIRENE"
```
tfardet, line 257: """

pynsee/utils/requests_session.py
```diff
@@ -0,0 +1,360 @@
...
+        return re.match(".*api-sirene.*", url)
+
+    def _set_sirene_key(self):
+
+        sirene_key = _get_credentials().get("sirene_key", None)
```
tfardet, line 262: None is unnecessary

pynsee/utils/requests_session.py
```diff
@@ -0,0 +1,360 @@
...
+        print_msg=True,
+    ):
+
+        headers = {"Accept": file_format}
+
+        try:
+            results = self.get(
+                url,
+                headers=headers,
+                verify=False,
```
tfardet, line 289: (why) do we need ``verify=False``

pynsee/utils/requests_session.py
```diff
@@ -0,0 +1,360 @@
...
+
+            if "status_code" not in dir(results):
+                success = False
+            elif code == 429:
+
+                if (
+                    os.environ.get(
+                        "PYNSEE_DISPLAY_ALL_WARNINGS", "false"
+                    ).lower()
+                    == "true"
```
tfardet, line 307: use a variable, you don't need the default argument in get since None != "true"

tfardet (changes requested):
Wow, even without the formatting, this is a big one, I did a 1st pass but it'll require another one for sure.

Anyway, though I think there are a few issues here and there, overall it looks much cleaner, thanks!

pynsee/localdata/get_old_city.py
```diff
@@ -37,7 +37,12 @@ def get_old_city(code, date=None):
     if date is not None:
         api_link = api_link + "?date=" + date
 
-    request = _request_insee(api_url=api_link, file_format="application/json")
+    # api_link = 'https://api.insee.fr/metadonnees/geo/commune/24259/precedents'
```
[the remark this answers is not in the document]
tfardet: Note: if you change this file, you could also cleanup lines 47-56 as in the other file? (using dataframe instead of concat)

pynsee/utils/_get_credentials.py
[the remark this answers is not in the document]
tgrandje: Better to keep the None value:  I seem to remember I wrote a snippet when a new connection is instantiated and this was triggered by non-None values.

pynsee/utils/_get_credentials.py
```diff
@@ -42,16 +46,33 @@ def _get_credentials() -> Dict[str, str]:
...
+            if key_dict["https_proxy"] != https_proxy:
+                logger.warning(
+                    f"Current https_proxy ({http_proxy}) is being overwritten "
+                    "by pynsee config. "
+                    "To reset it, please use pynsee.utils.init_conn.\n"
+                    f"Used http_proxy is now {key_dict['http_proxy']}"
+                )
+                https_proxy = key_dict["https_proxy"]
+
+            if not isinstance(http_proxy, str):
```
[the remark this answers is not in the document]
tgrandje: That'd be ok with me. Maybe we should drop the proxies being stored in the credentials altogether? What do you think?

pynsee/utils/_get_credentials.py
```diff
@@ -42,16 +46,33 @@ def _get_credentials() -> Dict[str, str]:
...
                 http_proxy = ""
-            if (https_proxy is None) or (not isinstance(https_proxy, str)):
+
+            if not isinstance(https_proxy, str):
                 https_proxy = ""
 
-            os.environ["http_proxy"] = http_proxy
-            os.environ["https_proxy"] = https_proxy
+            key_dict["https_proxy"] = https_proxy
+            key_dict["http_proxy"] = https_proxy
```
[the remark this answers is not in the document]
tgrandje: That sure could be enhanced, that part gave me a slight headache... I think this was to overwrite `key_dict["https_proxy"]` in case this is a None value... (And btw, there's a typo on http*s*_proxy...)

I'm definitely in favour of not storing the proxies on the disk, it ends up being a nightmare when working on a laptop...

pynsee/utils/init_conn.py
tgrandje: Yes, that would be the same anyway for sirene_key. Again, the aim here was to force an update of credentials when setting an empty string for http_proxies (= back to home network with my laptop) which is quite different from http_proxies not being set.

pynsee/utils/init_conn.py
tgrandje: I think so, and I agree with your assessment about the warnings. I'm not sure either what would be the best course and I thought that could be kept for later...

pynsee/utils/init_conn.py
tgrandje: That something I thought I had removed 😳 !

pynsee/utils/requests_session.py
tgrandje: No, the codes at the beginning are from the API portal's documentations. The other here are standard codes. But I can add those translations here of course.

pynsee/utils/requests_session.py
tgrandje: Mostly not to trigger an error in case of certificate expiration. That's a trouble I have also encountered here and there, so I choose not to remove it.

pynsee/utils/_get_credentials.py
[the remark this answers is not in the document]
tfardet: long-term (but starting as soon I this is done), I had planned to set up a config object that would store this and the credentials (now only the SIRENE key), as well as global session objects (rather than using globals here and there), so I would keep them there as I think it can be convenient and have seen people encounter issues with the proxies environment variables, but I don't have a very strong opinion about it so if you think it's not a good idea, we can also leave them out of it

pynsee/utils/_get_credentials.py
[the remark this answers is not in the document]
tfardet: ok, we can stick to environment variables for proxies, then, as far as I'm concerned
maybe just check with @hadrilec that he doesn't have any major issue with it and do it afterwards, though
I'll go through the code again tomorrow and will propose something

pynsee/utils/_get_credentials.py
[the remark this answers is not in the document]
tgrandje: I'd say we drop those for now.
This will help keeping the code being simpler (most of my troubles with init_conn, the session object and the credentials were about only the sirene_key being the trigger in _get_credentials). We can alter this afterwards.

pynsee/utils/_get_credentials.py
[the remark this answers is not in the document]
tgrandje: Same here, I'm done for this evening, I'll try to find some time tomorrow (or monday evening if not that). Thanks for your suggestions!

tfardet:
So I've been over the code again.
Unless I missed something, what we want is that the session object automatically check the environment variables upon creation and otherwise defaults to the stored values.
If so, this seems simple enough and we just need:

```python
def get_env_case_insensitive(x: str):
    "Fetch an environment variable (case insentitive)"
    val = os.environ.get(x)
    if val is None:
        return os.environ.get(x.upper())
    return val


class PynseeAPISession(requests.Session):

    def __init__(self):
        super().__init__()
        self._mount_adapters()
        
        credentials = _get_credentials()
        
        for k in ("http_proxy", "https_proxy", "sirene_key"):
            v = get_env_case_insensitive(k)
            
            if v:
                credentials[k] = v

        self.proxies.update({
            "http": credentials.get("http_proxy"),
            "https": credentials.get("https_proxy"),
        })

        username = os.environ.get("USERNAME", "username")
        version = pynsee.__version__

        # Note : geoplatform seems to impose the "/version" to the user-agent
        self.headers["User-Agent"] = f"python_pynsee_{username}/{version}"

        self.sirene_key = credentials.get("sirene_key")
        self.headers["X-INSEE-Api-Key-Integration"] = self.sirene_key
```

everything else uses the session object and it is updated every time (this will be a problem if we want to have global session objects, though, I'll think about it, but it's irrelevant for this PR

tgrandje:
Thanks, I'll check this later on (on my phone right now...).

Don't worry about the global session object, I already had to remove this in that other [PR](https://github.com/InseeFrLab/pynsee/pull/216) and I don't think we need it altogether. I'll work on updating that PR as soon as we finish this.

tgrandje:
> So I've been over the code again. Unless I missed something, what we want is that the session object automatically check the environment variables upon creation and otherwise defaults to the stored values. If so, this seems simple enough and we just need:
> 
> ```python
> class PynseeAPISession(requests.Session):
> 
>     def __init__(self):
>         super().__init__()
>     [...]
> ```
> 
> everything else uses the session object and it is updated every time [...]

This is not going to work when you instantiate a new connection with init_conn, is it? I think we still need to be able to configure the Session object with given arguments to override either those from `os.environ` or the config file...

pynsee/utils/init_conn.py
tgrandje: > str | None
Just found why I had'nt ever used that syntax... It appeared in python 3.10+ 😄

tfardet: Ow, sorry, thought it was 3.9... Might bad, we'll have to stick with ``Optional``, then

tgrandje:
Still not done with this, maybe tomorrow... I'd like to add some tests, to ensure we don't mess up with the order of environ/arguments/previous configs. Whatever we decide on this, this part is complex enough to have it's own automatic test!

tfardet:
> This is not going to work when you instantiate a new connection with `init_conn`, is it? I think we still need to be able to configure the Session object with given arguments to override either those from `os.environ` or the config file...

Right, forgot that use case!

```python
class PynseeAPISession(requests.Session):

    def __init__(
        self,
        sirene_key: Optional[str] = None,
        http_proxy: Optional[str] = None,
        https_proxy: Optional[str] = None
    ):
        super().__init__()
        self._mount_adapters()
        
        credentials = _get_credentials()
        
        for k in ("http_proxy", "https_proxy", "sirene_key"):
            v = get_env_case_insensitive(k)
            
            if v:
                credentials[k] = v

        self.proxies.update({
            "http": http_proxy or credentials.get("http_proxy"),
            "https": https_proxy or credentials.get("https_proxy"),
        })

        username = os.environ.get("USERNAME", "username")
        version = pynsee.__version__

        # Note : geoplatform seems to impose the "/version" to the user-agent
        self.headers["User-Agent"] = f"python_pynsee_{username}/{version}"

        self.sirene_key = sirene_key or credentials.get("sirene_key")
        self.headers["X-INSEE-Api-Key-Integration"] = self.sirene_key
```

I guess this should work?

tgrandje:
@tfardet I did a bunch of edits:
* simplify _get_credentials and rename it to a more expressive name: retrieve configuration **only** from the latest config file
* remove API call checkups from init_conn and transfer it to the new session object (allowing to use quick and dirty patches to emulate a valid connection to all APIs during tests)
* ensure any invalid SIRENE key is never stored to the config file, but do not block pynsee if this is invalid as you may want to queries other APIs inly
* transfer warnings of os environ variables being used to the new session object and decline these warnings for each config key
* prune dead code (_set_sirene_key method)
* add tests ensuring new session object's configuration in diverse settings (explicit arg, os.environ or previously saved config file)

Your suggestions were great, but those were only valid if we simplify _get_credentials to the bare minimum (retrieval from config file) and we run the warnings from the session object. (By the way : I'm not certain it is best to warn that pynsee is loaded with os environment variables instead of previous keys stored in an obscur config file, but I can live with it 😄 ...)

I will still loop around all your suggestion to check if there are still things amiss, but the process should already be much cleaner. Thanks again for your suggestions!

pynsee/utils/requests_session.py
tgrandje: You're right. Let's got for an exception instead, the code should be stopped anyway...

pynsee/utils/requests_session.py
tgrandje: My bad, that one's from me!

tgrandje:
Ok, that should be better. Maybe two more elements to check (I think), but that'll be for tomorrow :
* whether we could pass raise_if_not_ok directly to `self.request` from `self.get` and not handle it 3 different times 
* maybe ditch the RuntimeError in favour of RequestsException

pynsee/utils/requests_session.py
```diff
@@ -0,0 +1,601 @@
...
+        self.headers.update(useragent)
+
+        self.sirene_key = (
+            sirene_key or get_from_configfile_and_deactivate_warn("sirene_key")
+        )
+        self.headers["X-INSEE-Api-Key-Integration"] = self.sirene_key
+
+        for k in ("http_proxy", "https_proxy", "sirene_key"):
+            if getattr(self, f"warning_{k}"):
+                _warn_env_credentials(k)
```
tfardet, line 140:
```suggestion
        config = {
            "http_proxy": http_proxy,
            "https_proxy": https_proxy,
            "sirene_key": sirene_key
        }

        stored_config = _get_credentials_from_configfile()

        for k, v in config.values():

            venv = get_env_case_insensitive(k)

            if v is None:
                # Use env variable if it exist, otherwise stored value.
                # Note: v might be set to the empty string for proxy config,
                # which is **not** the same as None
                if venv is not None:
                    config[k] = venv
                    _warn_env_credentials(k)
                else:
                    config[k] = stored_config[k]

        self.proxies.update(
            {
                "http": config["http_proxy"],
                "https": config["https_proxy"],
            }
        )

        username = os.environ.get("USERNAME", "username")
        version = pynsee.__version__
        useragent = {"User-Agent": f"python_pynsee_{username}/{version}"}

        # Note : geoplatform seems to impose the "/version" to the user-agent
        self.headers.update(useragent)

        self.sirene_key = config["sirene_key"]
        self.headers["X-INSEE-Api-Key-Integration"] = self.sirene_key
```

tfardet (commented):
Noted that "" is to deactivate the previous value, sorry, I was a bit slow on this one...
I think I found a way to simplify the ``__init__`` of the session object, could you check if this indeed works?

pynsee/utils/requests_session.py
```diff
@@ -212,16 +224,53 @@ def kw_adapter(api: str):
...
             adapter = LimiterAdapter(**config["rates"], **kw_adapter(api))
             self.mount(config["url"], adapter)
 
     def request(
-        self, method, url, timeout=(10, 15), raise_if_not_ok=True, **kwargs
-    ):
+        self,
+        method: str,
+        url: str,
+        timeout: Union[tuple, int] = (10, 15),
```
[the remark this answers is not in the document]
tgrandje: Not sure about that. You could still use a simple integer for the timeout, it's authorized by requests.Session object.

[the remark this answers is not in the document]
tfardet: my bad, you're right!

pynsee/utils/requests_session.py
```diff
@@ -329,9 +483,7 @@ def _request_api_insee(
             elif code == 429:
 
                 if (
-                    os.environ.get(
-                        "PYNSEE_DISPLAY_ALL_WARNINGS", "false"
-                    ).lower()
+                    os.environ.get("PYNSEE_DISPLAY_ALL_WARNINGS").lower()
                     == "true"
```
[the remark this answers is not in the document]
tgrandje: No problem, I'll add that (I hadn't understood what you meant)

pynsee/utils/init_conn.py
```diff
@@ -123,13 +84,29 @@ def init_conn(
...
+        logger.info("Credentials have been saved.")
+
+    elif invalid_requests:
+        # some APIs are working, excepting SIRENE. User has already been warned
+        # about this.
+        config = {
+            "sirene_key": None,
+            "http_proxy": http_proxy,
+            "https_proxy": https_proxy,
+        }
```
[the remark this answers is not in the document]
tfardet: but then we would overwrite a potentially valid sirene_key with None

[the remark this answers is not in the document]
tgrandje: Good point, but what if the user if initiating a session without the sirene_key because he's not gonna use that API?
You will never store the proxies even if this was intended behaviour.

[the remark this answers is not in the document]
tfardet: good point too, we need to load the old sirene value from the config file and reset it to its old value instead of None if there is an issue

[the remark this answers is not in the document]
tfardet: I guess we can save if invalid request is empty or only contains sirene?

tgrandje:
@tfardet ok that's time we might be getting close. What do you think?

pynsee/utils/init_conn.py
```diff
@@ -84,30 +90,28 @@ def init_conn(
...
+            else:
+                logger.warning("Previous SIRENE key has been restored.")
+
+    config = {
+        "sirene_key": sirene_key,
+        "http_proxy": http_proxy,
+        "https_proxy": https_proxy,
+    }
+
+    if not invalid_requests:
```
tfardet, line 114: I thought you wanted to save if sirene was the only issue in the api?

If so, use ``if set(invalid_requests).issubset({"Sirene"})`` or ``if {"Sirene"}.issuperset(invalid_requests)``.

pynsee/utils/requests_session.py
```diff
@@ -89,37 +89,34 @@ def __init__(
...
-            return credentials.get(var)
+                if venv is not None:
+                    config[k] = venv
+                    _warn_env_credentials(k)
+                else:
+                    try:
+                        config[k] = stored_config[k]
+                    except KeyError:
+                        # no previous config stored
+                        pass
```
tfardet, line 114:
```suggestion
                    config[k] = stored_config.get(k, config[k])
```

tfardet (commented):
Just a few details left, I think the next round will be the last!

tfardet (approved):
Great, I'll merge that as soon as the tests pass!

PR #241, opened 2025-02-03
test xdist impact with loadscope

tgrandje:


tests/utils/test_pynsee_utils.py
```diff
@@ -127,7 +124,11 @@ def test_dummy_sirene_token_is_not_stored(self):
         with open(config_file, "w") as f:
             json.dump({"sirene_key": "spam"}, f)
 
-        init_conn(sirene_key="eggs")
+        init_conn(
+            sirene_key="eggs",
+            http_proxy=os.environ.get("http_proxy"),
+            https_proxy=os.environ.get("https_proxy"),
```
tfardet, line 130: this should not be necessary, the env variables are checked in the session object, aren't they?

tfardet:
BTW, I cancelled the workflows because they are not testing the new yml file

tests/utils/test_pynsee_utils.py
tgrandje: That's to make the tests work locally when you're behind a proxy.

PR #245, opened 2025-02-05
fix tests failing because of deactivation of retries

tgrandje:


tfardet:
I'm a bit confused about this, since I haven't spent as much time on it as you did:

- what makes you think that it's safe to keep the no-retry policy for the other tests?
- couldn't we just reduce the number of retries for this test rather than having all this patching?

tgrandje:
`patch_retries` was there only to speed up tests meant to end with a `RequestException`.

Now, it's only used in `test_init_conn_with_dummy_proxy` which is meant to fail due to a wrong proxy configuration . Let's say it's a proxy (!) test to check that proxies are correctly used (as nothing works if proxies are not configured correctly). So in that case, no https request is ever performed, so we don't need to be cautious about the API portal's  behaviour. (To be honest, I'm not even sure that errors on proxy configuration triggers retries, so maybe we could drop that patch altogether...).

About that last patching on `test_overriding_insee_config_and_environ4`: yes, we could either keep all retries or reduce their number. But :
* the test is meant to fail with a 400-something results on SIRENE calls, which should trigger retries. It's long as tests goes (granted, it should still be only one minute...) and I think we had troubles enough with long tests (more so that we haven't reestablished all of them yet) ;
* I'm not sure keeping some retries (let's say 3 retries for 3seconds) would be reliable anyway : we already had to expand the number of retries from 3 to 7 to have a consistent behaviour for other tests anyway...

tgrandje:
> (To be honest, I'm not even sure that errors on proxy configuration triggers retries, so maybe we could drop that patch altogether...).

Just tested it without : I confirm we still need this. And it certainly is slower that what I intended from urlib3's retry state's (2min30 to run all tests in my machine with the original retries instead of 5secs with the patch).

tfardet (approved):

PR #250, opened 2025-02-12
Fix poverty examples

tgrandje:
Fix both poverty examples


tgrandje:
Just cancelled the tests as those are not testing any of the changed files.

tfardet (approved):

PR #251, opened 2025-02-12
Update documentation

tfardet:
This PR:

- cleans up the requirements for the documentation
- includes the version in the config file
- switches to Napoleon since a) I like it better and b) we have both google and numpy style docstrings
-  adds intersphinx mapping to ``pandas``, ``geopandas``, and ``requests``
- adds a module page (the only one that is indexed) with all functions and classes
- documents (some of?) the environment variables
- removes quarto and m2r2 stuff
- ensures that references to functions/classes link to their documentation


tfardet:
Not running the tests since I'm not touching code files.
@hadrilec is there a way to test the doc in advance on RtD? My handle there is ``TFardet`` if you can add me as a maintainer

PR #254, opened 2025-02-16
Update api_subscription.rst

milena-git:
pynsee.utils.init_conn > pynsee.utils.init_connection

Pour que l'exemple fonctionne il me semble


hadrilec:
thanks milena!

tfardet:
thanks @milena-git !

Side note for the future: we started reducing name shadowing issues with @tgrandje by having file names and functions differ.
This is done for ``init_conn``, so it is no longer necessary to do ``from pynsee.utils.init_connection import init_conn`` but ``from pynsee.utils import init_conn`` is enough

PR #259, opened 2025-03-04
Support parquet files in ``download_file``

### LINKED ISSUE #258: Issue with `pynsee.download.download_file` on latest commuter data in `parquet` format
Hello
It seems that the latest commuter (Déplacement Domicile-Travail) data `RP_MOBPRO_2021` and `RP_MOBSCO_2021` in `parquet` format generates an error with `pynsee.download.download_file` method.
Example below:

```python
from pynsee.download import download_file
df = download_file('RP_MOBSCO_2021')
```

The download seems to work but we get the error message below before we can get the output:
"**UnboundLocalError: cannot access local variable 'df_insee' where it is not associated with a value**"

Any idea of how to fix the issue?
Many thanks

tfardet:
Fixes #258

### DIFF
--- a/pynsee/download/_load_data_from_schema.py
+++ b/pynsee/download/_load_data_from_schema.py
@@ -40,6 +40,7 @@ def _load_data_from_schema(
     Returns:
         pd.DataFrame -- The required dataset is returned as pd.DataFrame object
     """
+    df_insee = None
     file_to_import = telechargementFichier["file_to_import"]
 
     if telechargementFichier["result"]["zip"] is True:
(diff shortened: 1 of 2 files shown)


hadrilec:
it tried, it works well, we can merge

daniel-odc:
Many thanks for your quick fix!
Has the new package version been released? Under which version number? If not, any idea of the release date?

tfardet:
> Many thanks for your quick fix!

My pleasure.

>  Has the new package version been released? Under which version number? If not, any idea of the release date?

We will try to make a bugfix release in the week-end or early next week, but we make no promises.
In the meantime, you can install the latest version from the git repository via: ``pip install pynsee@git+https://github.com/InseeFrLab/pynsee``

PR #261, opened 2025-03-06
expose silent arg in get_geodata

tgrandje:
This exposes the `silent` arg in `get_geodata` (it was only accessible in the inner `_get_geodata` function) and pass it to the private function.

Also fix snake case for `crs_polygon`.

### DIFF
--- a/pynsee/geodata/get_geodata.py
+++ b/pynsee/geodata/get_geodata.py
@@ -13,6 +13,7 @@ def get_geodata(
     update: bool = False,
     crs: Any = "EPSG:3857",
     constrain_area: Optional[GeoDataFrame] = None,
+    silent: bool = False,
 ) -> GeoFrDataFrame:
     """Get geographical data with identifier and from IGN API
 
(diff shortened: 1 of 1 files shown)


pynsee/geodata/get_geodata.py
```diff
@@ -25,6 +26,8 @@ def get_geodata(
 
         constrain_area (:class:`~geopandas.GeoDataFrame`, optional): GeoDataFrame used to constrain the area of interest. Defaults to None.
 
+        silence (bool, optional): whether to print warnings or not. Defaults to False.
+
```
tfardet, line 30: could you add a ``.. versionadded:: 0.3.0`` with "added `silent` parameter"?

tgrandje: Sure, sorry I have not the reflex to log docstrings (yet). Are we certain we have enough changes to justify a jump to 0.3 version (this seems very minor to me).

tfardet: I just assumed we would not necessarily tag a new version until we have enough changes, so this would become 0.3.0...

tgrandje: Noted. Though the geodata PR from this morning might be enough to justify a new version. I still don't know if this is a frequent behaviour, but that was troubling enough...

Issue #213, opened 2024-08-30
Melodi - new insee API = mew module


hadrilec:
A new module should be developed to tap into the brand new resources made available by INSEE
https://portail-api.insee.fr/catalog/api/a890b735-159c-4c91-90b7-35159c7c9126/doc?page=ee625968-272a-4637-a259-68272aa63766

hadrilec:
WIP PR created #217

tfardet:
@hadrilec we did not have time to check the PR that @linogaliana made and merged, should we take some time to review it at some point and see whether some of the things you planned to add in your PR should still be added?

I haven't yet managed to dig into Melodi but I think it should provide us with data that still seems to be missing from pynsee.

For instance, a previous issue I mentioned about tourism data still stands: the latest data is ``GEO2020TOUR2021`` but much more recent data is available and seemed to be linked to Melodi: https://www.insee.fr/fr/statistiques/2021703

hadrilec:
it is definitely worth having a look at the PR from Lino, that being said, I think that for Melodi we should try to start from a blank page with a brand new module, and then slowly decommission localdata module and other deprecated data flows

tfardet:
ok, tu aurais moyen de merge master sur ta PR, que je puisse y jeter un coup d'oeil, du coup ?

PR #262, opened 2025-03-06
Bugfix/examples

tgrandje:
Update some examples from the doc:

* remove geofrdataframes and transform
* update code with latest numpy/pandas versions
* remove outdated references to API keys
* use pynsee to download latest IRIS dataset with built-in functions
* refacto a bit of code (mostly due to the IRIS dataset's slight changes)
* update some colormaps to be more "colorblind-friendly" and switch to monochrome palettes when deemed appropriate* 

\* that being said: _de gustibus_ and so on... : so if you disagree with the new palettes, no problem for setting it back!


tgrandje:
Note that deaths/births example raises (on cell #5) :
> [...]\Lib\site-packages\pandas\core\indexes\base.py:7588: FutureWarning: Dtype inference on a pandas object (Series, Index, ExtensionArray) is deprecated. The Index constructor will keep the original dtype in the future. Call `infer_objects` on the result to get the old behavior.

tfardet:
> Ah, population_pyramid is ko without any prospect of resetting it right away. It used localdata API which has been discontinued (you are supposed to switch to MELODI). I think we should dig up #213.

yes, this definitely needs doing but I've no idea when I might be able to work on this... unlikely any time before the end of the year.


> I also think the test failure is due to localdata's / melody discrepancy.

by test failure, you mean ``test_get_local_data_latest`` or a test related to ``population_pyramid``?


> @tfardet what do you think we do on this? Tag the example as obsolete and keep it in the doc for the moment ? Maybe even double check and disable the failing test on `test_pynsee_localdata` ?

Tagging the example as obsolete sounds good until we can work on updating it.
No idea about the failing localdata tests, though...


> The missing dataset on localdata seems to be the census dataset (dataset_version being `GEOlatestRPlatest`). I think the newest dataset is [that one](https://catalogue-donnees.insee.fr/fr/catalogue/recherche/DS_RP_POPULATION_COMP)...

Yeah, this looks weird, probably the people from INSEE removed some data because this should not be able to fail.
That said, GEOlatestRPlatest is also probably a situation where we should use Melodi...

PR #264, opened 2025-03-07
fix/ci-release

tgrandje:
Should avoid creating a release which already exists.

@tfardet any idea how to test this?

Also : `actions/create-release` and `actions/upload-release-asset` are not maintained anymore , `pypa/gh-action-pypi-publish` seems outdated. As I'm never quite sure how to test CIs, I'd say we could start with fixing the limited PR as it is?

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -46,8 +46,14 @@ jobs:
           name: artifact
           path: dist
 
+      - name: Check if Release Exists
+        id: check_release
+        run: |
+          if gh release view ${{ github.ref }}; then echo "exists=true" >> $GITHUB_ENV; else echo "exists=false" >> $GITHUB_ENV; fi
+
       - name: Create GitHub Release
         id: create_release
+        if: env.exists == 'false'
         uses: actions/create-release@v1
         env:
           # This token is provided by Actions, you do not need to create your own token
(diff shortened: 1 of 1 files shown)


tfardet:
honestly, I don't know how to trigger this workflow except by creating the release, so I don't think we will ever encounter ``exists=false``, so unless you know how to make this case happen, I'd just remove the github release part altogether

tgrandje:
Sure!

As I was not certain why this was there in the first place I let it and added the condition ; I agree it makes more sense to remove this if possible. @hadrilec do you have any insights on this?

hadrilec:
I used to test the ci by uploading to pypitest
```
- name: Upload Release Assets to PyPI
        uses: pypa/gh-action-pypi-publish@v1.4.2
        with:
          user: __token__
          password: ${{ secrets.pypi_password }}
          skip_existing: true
          repository_url: https://test.pypi.org/legacy/

```

tgrandje:
Yet that won't work until this is pushed to master (I think). Maybe we keep this option for latter tests if we choose to move away from those unmaintained actions?

tfardet (approved):

tfardet:
@hadrilec this seems OK to me, but since you were the one doing the releases before, I just want to double-check that this is fine with you and that we did not miss something.

What I usually do is :

1. bump the version
2. make a release associated to the commit with the bumped version
3.  let the CI run for the new release and make the push to PyPi

as far as I can tell, the changes done here are now compatible with this workflow (previously the CI config files would attempt to re-create a release, which would fail) but maybe you were doing things differently.

PR #267, opened 2025-03-09
chore: update coverage and cleanup badges

tfardet:
For some reason the tests and coverage badges were taken from static files in the directory.
I removed this, (hopefully) fixed the coverage reporting and updated the badges


.github/workflows/pkgTests.yml
```diff
@@ -46,9 +45,9 @@ jobs:
       env:
         sirene_key: ${{ secrets.SIRENE_KEY }}
       run: |
-        pytest -v --cov -n auto
+        pytest -v --cov --cov-report=xml -n auto
```
tgrandje, line 48: I have not run this command (yet), but from previous experience, this might include coverage on tests themselves. I had to run ` run pytest --cov-config pyproject.toml --cov`, and in the pyproject.toml add the following config:
```
[tool.coverage.run]
omit = [
    "tests/*",
    ]
```

Can you double check on that?

README.md
```diff
@@ -1,15 +1,12 @@
...
-[![](https://readthedocs.org/projects/pynsee/badge/?version=latest)](https://pynsee.readthedocs.io/en/latest/?badge=latest)
-[![](https://img.shields.io/badge/python-3.9%20%7C%203.10%20%7C%203.11%20%7C%203.12-blue.svg)](https://www.python.org/)
-[![](https://img.shields.io/badge/code%20style-black-000000.svg)](https://pypi.org/project/black/)
-[![](https://img.shields.io/pypi/dm/pynsee)](https://pypistats.org/packages/pynsee)
+[![PyPi version](https://badge.fury.io/py/pynsee.svg)](https://pypi.org/project/pynsee/)
+[![Tests status](https://github.com/InseeFrLab/pynsee/actions/workflows/pkgTests.yml/badge.svg)](https://github.com/InseeFrLab/pynsee/actions)
+[![Coverage status](https://codecov.io/gh/InseeFrLab/pynsee/branch/master/graph/badge.svg?token=TO96FMWRHK)](https://codecov.io/gh/InseeFrLab/pynsee?branch=master)
+
+[![Documentation](https://readthedocs.org/projects/pynsee/badge/?version=latest)](https://pynsee.readthedocs.io/en/latest/?badge=latest)
+[![Supported python versions](https://img.shields.io/badge/python-3.9%20%7C%203.10%20%7C%203.11%20%7C%203.12-blue.svg)](https://www.python.org/)
```
tgrandje, line 7: I thought we could simplify it this way from another project:  `[Supported Python Versions](https://www.python.org/)` but it returns only python 3. One thing leading to another, I think this may be due to 3.9+ being the only reference in the pyproject.toml. Shouldn't we be coherent between that and the badge?

coverage.xml
tgrandje, line 1: Do we add this to gitignore (in case of local tests being run with copy/paste commands from the CI)?

.github/workflows/pkgTests_pull_requests.yml
```diff
@@ -48,9 +48,9 @@ jobs:
       env:
         sirene_key: ${{ secrets.SIRENE_KEY }}
       run: |
-        pytest -v --cov -n auto
+        pytest -v --cov --cov-report=xml -n auto
```
tgrandje, line 51: same thing on this

hadrilec:
guys, shall we merge this one?

README.md
tfardet: what do you propose?

Issue #256, opened 2025-02-18
metadata download in macrodata module should not retrieve metadata on all datasets but only on one


hadrilec:
the metadata download is too slow

tgrandje:
I just timed _get_insee for a start (I hope I didn't misunderstood you : you *were* speaking of `macrodata.get_dataset`, isn't it? If that's the case, I think the metadata part is only a small element of what's "too slow").

We will need to refactorized _get_insee first: what's taking forever seems to be the parsing of the xml file. I think that's mostly due to the `getElementsByTagName` and `DataFrame` construction/concatenation in the inner loop. We should be able to speed up things a lot with this.

I'll start working on that.

PR #268, opened 2025-03-13
Feat/fasten macrodata get dataset

tgrandje:
Working on https://github.com/InseeFrLab/pynsee/issues/256

For now this mostly speeds `_get_insee` up (~90% faster on parsing the xml, download excepted).

This is currently using `lxml` instead of `xml.dom.minidom` due to :

* me knowing only the first one 
* and `lxml` having a reputation for being the fastest available.

I'll double check on that and see if this is pertinent to addd `lxml` as a new dependency or if we can stick to the standard libraries.


tgrandje:
I've also removed some dataframes construction from an inner loop in `def _get_dataset_dimension`.

tgrandje:
Ok, this is as good as I should be. I tested this on `get_dataset("IPC-2015", filter="M......ENSEMBLE...CVS.2015",         updatedAfter="2017-07-11T08:45:00")` with and without metadata. Execution times varies due to download, but I got (mostly) consistent timeranges:

* with metadata :
    * previous code:
         * CPU times: total: 1min 53s
         * Wall time: 2min 1s
    * newest code:
         * CPU times: total: 10.1 s
         * Wall time: 19.6 s
* without metadata:
    * previous code:
         * CPU times: total: 1min 20s
         * Wall time: 1min 28s
    * newest code:
         * CPU times: total: 3.39 s
         * Wall time: 10.4 s

I've also :
* fixed the cascading usage of `update` (mostly for metadata retrieval)
* removed the testing of idbank correspondence file referring to "future" dates (I've also tested scraping the correct file directly from the info webpage but this is not faster as it requires at least two http requests)
* removed the `pynsee_idbank_loop_url` os variable as I don't think it was used, even in commented out tests (@hadrilec do you have any insights on why this was here in the first place?)

(And I'm back to standard libraries, regarding to my previous comment on the draft PR.)

hadrilec:
about pynsee_idbank_loop_url, not sure I remember well, I guess this legacy was meant for testing purposes, indeed it should have been removed, great that you did it!

hadrilec:
well, done for the better performances! I understand that the downloaded metadata is now only related to one dataset?

tgrandje:
I'm not sure of this.

In fact, it might have already been fixed by @tfardet in previous commits: I tested it on `get_dataset("IPC-2015", filter="M......ENSEMBLE...CVS.2015", updatedAfter="2017-07-11T08:45:00")` and the only metadata gathered were the correct ones.

Just in case, I upped the check in `_get_metadata` [there](https://github.com/InseeFrLab/pynsee/blob/a6a97088475ca1e2a81ab9e148fa8392b00fbfd8/pynsee/macrodata/_get_metadata.py#L98-L99). Do you think it covers this issue?

(If perchance you have the specific code to reproduce this issue, send it to me and I'll check later on.)

pynsee/macrodata/_get_insee.py
```diff
@@ -1,131 +1,129 @@
...
 
 
+def _set_date(df: pd.DataFrame) -> None:
+    """
+    Parse dates on a dataframe issued by macrodata module.
+
+    Adds a "DATE" column (computed from the "TIME_PERIOD" original column).
+    The dataframe is changed in place.
+
+    Internal note: leave this as a distinctive function to to improve tests
```
tfardet, line 24:
```suggestion
    Internal note: leave this as a separate function to to improve tests

```

pynsee/macrodata/_get_insee.py
```diff
@@ -1,131 +1,129 @@
...
-    data_final = pd.concat(list_series)
+    # place DATE column in the first position
+    data = data[["DATE"] + [c for c in data if c not in ["DATE"]]]
 
-    # index and sort dataframe by date
-    # data_final = data_final.set_index('DATE')
-    if "DATE" in data_final.columns:
-        data_final = data_final.sort_values(["IDBANK", "DATE"])
+    if "DATE" in data.columns:
+        data = data.sort_values(["IDBANK", "DATE"]).reset_index(drop=True)
```
tfardet, line 111: move to ``_set_date`` and make it return the df?

pynsee/macrodata/_get_metadata.py
```diff
@@ -94,10 +94,10 @@ def _get_dataset_metadata_core(dataset, update=False, silent=False):
         [idbank_list_dataset, df_cleflow_splitted], axis=1
     )
 
-    n_dimensions = len(dataset_dimension.index)
+    for irow, dim_id in dataset_dimension["dimension"].items():
+        if dim_id not in idbank_list_dataset.columns:
+            continue
```
tfardet, line 99: was there an issue that required this? if so, what was it and on which dataset?

tests/macrodata/test_pynsee_macrodata.py
```diff
@@ -116,7 +117,11 @@ def test_get_date(self):
             "001580394",
         )
         assert isinstance(data, pd.DataFrame)
-        assert _get_date(freq="TEST", time_period=3) == 3
+        df = pd.DataFrame([[3, "blah"]], columns=["TIME_PERIOD", "FREQ"])
+        _set_date(df)
+        assert (
+            df["DATE"].values == df["TIME_PERIOD"].values == np.array(3)
+        ).all()
```
tfardet, line 124: if you have time, could you add actual tests for the different Q/S/B variants?

tfardet (changes requested):
Thanks for all that work!
I made a few suggestions, the additional tests are optional; it would be nice if you have time, but I would not let it block the PR.

pynsee/macrodata/_download_idbanks.py
```diff
@@ -36,13 +50,28 @@ def _download_idbank_list(
...
+    data : pd.DataFrame
+        correspondance between idbank and it's dimensions as DataFrame
+    """
+
     todays_date = date.today()
 
     main_link_en = "https://www.insee.fr/en/statistiques/fichier/2868055/"
     main_link_fr = "https://www.insee.fr/fr/statistiques/fichier/2862759/"
 
+    patterns = []
```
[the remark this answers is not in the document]
tgrandje: no clue, good catch !

pynsee/macrodata/_download_idbanks.py
```diff
@@ -51,62 +80,68 @@ def _dwn_idbank_files():
...
-
-    with PynseeAPISession() as session:
-        try:
-            data = _dwn_idbank_file(file_to_dwn=file_to_dwn, session=session)
-            idbank_file_not_found = False
-        except Exception:
-            idbank_file_not_found = True
+        pass
+    else:
+        with PynseeAPISession() as session:
```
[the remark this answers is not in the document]
tgrandje: Well, as a matter of fact, no : it doesn't speed up the code (maybe because it's only one line of code and a dict)😄  But anyway you're right that this could be more concise !

tests/macrodata/test_pynsee_macrodata.py
tgrandje: I just checked: in fact, `get_series("001694056", "001691912", "001580062", "001688370", "010565692", "001580394")` returns a dataframe with all variations on `"TIME_PERIOD"`. That's the reason there was only a dummy test on `_set_date`.

pynsee/macrodata/_get_metadata.py
tfardet: I'm not sure I got everything... do you mean that you don't know whether this is required or not?

tgrandje: No, I couldn't isolate a specific case (that being said, I didn't check thorougly). But as #256 's title is _metadata download in macrodata module should not retrieve metadata on all datasets but only on one_, I figure this happened indeed. Maybe that was in the past.

What's for sure right now, is that the check already exists, but at the end of the loop execution here: https://github.com/InseeFrLab/pynsee/blob/31f0096b687e3f2282136911a143d2404cdf373a/pynsee/macrodata/_get_metadata.py#L118. So even if this is a legacy security, we should bring it earlier.

tfardet (approved):

Issue #271, opened 2025-03-27
download tag_com_2025 raise ValueError


tgrandje:
`"AIRE_URBAINE"` has been removed from DoReMiFaSol and replaced by `"TAG_COM_XXXX"` where XXXX stands for vintage.

Right now, TAG_COM_2005 raises a ValueError, most certainly due to an old excel being used to create the file and the stylesheet.

tfardet:
Fixed in #272

PR #272, opened 2025-03-27
Bug/fix ValueError on non-aRGB-hex-values in excel files

tgrandje:
https://github.com/InseeFrLab/pynsee/issues/271

### DIFF
--- a/pynsee/download/_download_store_file.py
+++ b/pynsee/download/_download_store_file.py
@@ -1,13 +1,18 @@
-import hashlib
-import warnings
 import difflib
+import hashlib
+import logging
 import os
+import re
 import tempfile
+import warnings
 
 from pynsee.download._download_pb import _download_pb
 from pynsee.download._import_options import _import_options
 from pynsee.download._get_dict_data_source import _get_dict_data_source
 from pynsee.download._check_url import _check_url
+
+
+logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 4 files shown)


tgrandje:
This PR:

* fixes the download of excel files containing styles not compatible with `openpyxl`
* and also allows to query a dataset with a custom "latest" TAG (for instance `"TAG_COM_..."` only exists as a vintaged dataset). That should add a bit of stability.

pynsee/download/_download_store_file.py
```diff
@@ -28,11 +33,27 @@ def _download_store_file(tempdir: tempfile.TemporaryDirectory, id: str):
     """
 
     dict_data_source = _get_dict_data_source()
+    caract = None
     if id in dict_data_source.keys():
         caract = dict_data_source[id]
     else:
-        suggestions = difflib.get_close_matches(id, dict_data_source.keys())
 
+        if id.rsplit("_", 1)[-1].lower() == "latest":
```
tfardet, line 41: switch to elif

tfardet: maybe adding an explanatory comment would be good, I understand it from the PR topic, but out of context it might have taken a while

tfardet (changes requested):
Looks good overall, thanks for the fix!
I just had some minor comments and then we should merge the fix (once it is merged you can update your previous PR where you removed the AIRE_URBAINE test)

tgrandje:
Done, thanks for your insights!

PR #275, opened 2025-05-11
fix: incorrect test in get_column_metadata + cleanup

tfardet:


hadrilec:
@tfardet on which dataset did you test the proposed changes?
I dont get what is the difference with the previous outcome, ie `get_column_metadata("RP_LOGEMENT_2016")`

tfardet:
@hadrilec the issue was on line 33 (now 35): the test did not test the right thing. I have no idea on which dataset it is supposed to be relevant, though...

hadrilec:
I checked, the code runs smoothly

PR #283, opened 2025-08-07
remove unnecessary warnings related to sirene key

hadrilec:


hadrilec:
@tfardet sorry it is late and I clicked too quickly again... I have just added an argument to the object PynseeApiSession to make sure missing credentials warnings are not triggered whenever the Sirene API is not used, tests passed

PR #300, opened 2026-03-26
Faster tests

tgrandje:
https://github.com/InseeFrLab/pynsee/issues/299 and https://github.com/InseeFrLab/pynsee/issues/296


tgrandje:
OK, this is good for review.

I dumped most of the tree decision managing department's retrieval as it was not easy to maintain and could somehow be quite slow. It's been simplified to 3 cases:

* either we're working on the department's dataset (same as before)
* or we're on a dataset where INSEE (now) displays insee dpt codes (that's the case for EPCIs for instance) ;
* or in any other case, it's fast enough to use a sjoin (assuming we already stored a preprocessed departments' geodataset).

During my manual tests, I've mostly seen awkward results for "collectivités territoriales" (Saint-Martin, Saint-Barthélémy) or some other exotic cases : for instance arrondissements municipaux datasets, cantons, where there are holes in french territory or when you choose to zoom in on Paris' inner suburbs and your current dataset has geometries maching Paris's external suburbs. I don't think that was ever perfectly rendered in pynsee...

pynsee/geodata/translate_and_zoom.py
```diff
@@ -10,12 +11,55 @@
...
+    dep = dep.set_geometry("insee_dep_geometry")
+
+    x = dep.sjoin(dep, how="left", predicate="overlaps").query(
+        "~code_insee_du_departement_right.isnull()"
+    )
+    if not x.empty:
+        # Geometries overlap:
+        # Force coverage validity (non-overlapping, edge-matched polygons)
+        # (This shouldn't be necessary with modern ADMINEXPRESS datasets and
+        # is here as a backup safe)
```
tfardet: move above the sjoin?

pynsee/geodata/translate_and_zoom.py
```diff
@@ -10,12 +11,55 @@
...
+        dep["insee_dep_geometry"] = dep.simplify_coverage(1)
+    dep["insee_dep_geometry"] = dep["insee_dep_geometry"].buffer(-1)
```
tfardet, line 59: shouldn't this occur only in the if? (otherwise there is no overlap so I don't see why we should change the correct polygons

pynsee/geodata/translate_and_zoom.py
```diff
@@ -74,7 +118,46 @@ def transform_overseas(
...
         gdf["insee_dep_geometry"] = gdf["geometry"]
     else:
-        gdf = _add_insee_dep(gdf.copy()).reset_index(drop=True)
+
+        # retrieve safe deps geometries
+        dep_cov = _deps_with_valid_coverage()
+
+        # detect INSEE department's codes
+        # (available in some of IGN's geodatasets, but with different patterns)
+        pattern = "codes?.*?insee.*?departement"
```
tfardet, line 132: could you list the valid options in the comment above?

pynsee/geodata/translate_and_zoom.py
```diff
@@ -74,7 +118,46 @@ def transform_overseas(
...
+        dep = gdf.columns[gdf.columns.str.match(pattern, flags=re.IGNORECASE)]
+        if not dep.empty:
+            dep = dep[0]
+            deps = gdf[dep].str.extractall("([0-9AB]{2,3})")
+            deps.index = deps.index.droplevel(-1)
+            deps = deps.loc[~deps.index.duplicated(keep="first")]
+            gdf["code_insee_du_departement"] = deps[0]
+
+        else:
+
```
tfardet: remove blank line

pynsee/geodata/translate_and_zoom.py
```diff
@@ -74,7 +118,46 @@ def transform_overseas(
...
+            # retrieve department's codes using a spatial join. The negative
+            # buffer on deps should be safe to be used without any duplication
+            # of gdf, except when there is a valid overlapping (ie regions
+            # covering multiple deps, interdep epcis...).
```
tfardet, line 145: I don't think the buffer part is relevant here anymore, is it?

pynsee/geodata/translate_and_zoom.py
```diff
@@ -74,7 +118,46 @@ def transform_overseas(
...
+            gdf = gdf.loc[~gdf.index.duplicated(keep="first")]
+            gdf = gdf.drop("index_right", axis=1)
+
+        gdf["code_insee_du_departement"] = gdf[
+            "code_insee_du_departement"
+        ].fillna("NR")
+
+        # Retrieve simplified geometries for deps. Note that it used a negative
+        # buffer (10 meters) which should not alter geographic transformations
+        # given it's range
```
tfardet: adapt to "may" if you indeed moved the buffer into the if

tfardet (changes requested):

pynsee/geodata/translate_and_zoom.py
tgrandje: No, the sjoin is merely a checkup to verify is this is an overlapping dataset : this used to be true in older datasets and should be ok (as it is the case right now), but there is no guaranty this is gonna stay the same.

pynsee/geodata/translate_and_zoom.py
tgrandje: No : if a geometry fully covers a ~city~ department, it will sjoin on a city and it's neighbours. The buffer of -1meter is gonna fix this in the overwhelming majority of cases.

EDIT : I meant a department and not a city, obviously... 😳

pynsee/geodata/translate_and_zoom.py
tfardet: sorry, I meant move the docstring

pynsee/geodata/translate_and_zoom.py
tfardet: sorry, I'm not getting this, could you maybe draw it on a piece of paper and take a photo? (or we can schedule a meeting to discuss it in real time if it's easier, in which case you can just email me)
To prepare for that discussion, even though I don't get the issue, I'm wondering why you chose not to test for small fractions of the original areas instead of modifying the original geometries.

tgrandje: Image

tgrandje: https://excalidraw.com/#json=sUTRvCvLrtHnFMbbEKmxF,O1CsByZYwfowqnS-xUqb7A

tgrandje: Is this clearer?

EDIT
The negative buffer helps avoiding case 1A. Problems might occur on case 2B, but that seems too exotic a case to be handled by pynsee (I think this will concern only points in a 2meters-large bandwidth around DROM territories).

pynsee/geodata/translate_and_zoom.py
tgrandje: This should still be used to avoid case 1A of the picture above.

pynsee/geodata/translate_and_zoom.py
tgrandje: This is still useful for case 1A (see above).

pynsee/geodata/translate_and_zoom.py
tgrandje: I fxed the comment (not a 10 meter range, but 1 meter)

pynsee/geodata/translate_and_zoom.py
tgrandje: > I'm wondering why you chose not to test for small fractions of the original areas instead of modifying the original geometries.

I'm not sure I understood what you meant either. Let's discuss this in our matrix chanel?

tgrandje:
Note for memory's sake: we also could detect multipoint datasets and link it to parent datasets. For instance `chef_lieu_d_epci` has a `lien_vers_l_epci` fields which links to `cleabs` in the `epci` dataset. But:

* that seems overly complex to maintain (it's a rare use case I think) while sjoin are fasts on a point vs. polygon match,
* and users may use the `LATEST` dataset vintage and there is no certitude that `LATEST:chef_lieu_d_epci` and `LATEST:epci` are of the same vintage.

PR #306, opened 2026-04-07
melodi

tgrandje:


tgrandje:
@hadrilec @linogaliana @tfardet @RLesur do any of you guys see the point in covering MELODI's "référentiel" endpoint through `pynsee`?

It doesn't seem to be used in the [R package](https://github.com/InseeFrLab/melodi/) and I have no usecase in my mind right now (that and the fact that the json returned seem overly complex to be used in a generic way)...

tgrandje:
Note: 2 tickets have been opened on INSEE's API portal.

* one regarding problems with pagination (for instance querying `DS_POPULATIONS_REFERENCE` with no filters). I don't think any workaround can be done for this in `pynsee`.
* one regarding `X-INSEE-Api-Key-Integration` header were MELODI seems to confuse the header with the one used for SIRENE consumption. I've added a fix in pynsee in the `PynseeAPISession` object.

hadrilec:
Hi, I had started in the branch catalog to write a function retrieving the list of datasets available via melodi. On the long term, I think it makes sense, as I understand more INSEE datasets shall be available melodi. I understand that DoReMiFaSol already uses melodi: https://github.com/InseeFrLab/DoReMIFaSol

tgrandje:
Hi @hadrilec , I've already re-used your code for the catalog (which I also think is gonna be madatory). I was instead speaking of the "référentiel" endpoint and I haven't quite seen the point (yet ?).

Right now, you got catalog (as already said), you got all available dimensions and values for a given dataset using the "range" endpoint : it seems to me that's enough ?

PR #308, opened 2026-06-15
feat(melodi): tests, docstrings, fix measures parsing

GrasClement:
Draft contribution aux JDOS : cible la branche #306
- Fix `_parse_dataset_observations` (voir [package R Melodi](https://github.com/InseeFrLab/melodi))
- Docstrings manquantes
- Tests des 4 fonctions publiques


GrasClement:
Hi, 

The JDOS are done, here is a quick closure summary for this PR, 

#### Bug Fixes
- Fixed hardcoded measure key in `_parse_dataset_observations` 
- Added guard for observations without "measures" field (prevents crashes)
- Corrected language filter logic in `get_idbank` 
- Improved handling of non-MELODI idbanks → returns empty DataFrame

#### New Content
- Docstrings for 4 public MELODI functions + 2 internal helpers
- Tests: 14 tests covering parsing, edge cases, and all public functions
- Example notebook: `example_internet_usage_melodi.ipynb`
- Updated MELODI docs integration and DDL deprecation warning

#### Code Maintenance
- Added `update` and `silent` parameters to all 4 public functions
- Updated imports and version (0.2.6)
- Updated `.gitignore` for development files

Cheers

tfardet:
Thanks @GrasClement !
@tgrandje may I let you have a look at this since it targets your previous PR?
I'm not sure I'll have the time to look at this any time soon...

pynsee/melodi/data.py
```diff
@@ -67,16 +112,26 @@ def _parse_dataset_observations(response: requests.Response):
             )
 
     if "measures" in obs.columns:
-        measures = obs.measures.str["OBS_VALUE_NIVEAU"].str["value"]
-        obs["OBS_VALUE_NIVEAU"] = measures
-        obs = obs.drop("measures", axis=1)
+        # déplie toutes les clés de mesure dynamiquement
+        # (le nom varie selon le dataset : OBS_VALUE_NIVEAU, OBS_VALUE_INDICE_DE_PRIX, etc.)
```
tgrandje, line 116: Should we switch the comments to english?

pynsee/melodi/data.py
```diff
@@ -67,16 +112,26 @@ def _parse_dataset_observations(response: requests.Response):
...
+        measures_expanded = pd.DataFrame(
+            obs["measures"]
+            .apply(lambda m: {k: v.get("value") for k, v in m.items()})
+            .tolist()
+        )
```
tgrandje, line 121: Isn't the key unique in a dataset? If that's the case maybe we could avoid the whole apply process, what do you think?

pynsee/melodi/data.py
```diff
@@ -664,8 +764,9 @@ def get_idbank(
     # test = get_range("DS_TICM_PRATIQUES", include_values=True)
```
tgrandje, line 764: While you're at it, would you mind removing the tailing tests under the `if __name__ == "__main__"` block?

tests/melodi/test_pynsee_melodi.py
```diff
@@ -0,0 +1,261 @@
...
+
+from pynsee.melodi import get_catalog, get_dataset, get_idbank, get_range
+from pynsee.melodi.data import _parse_dataset_observations, _parse_metadata
+
+# ---------------------------------------------------------------------------
+# Mocks: the only mocks in the pynsee tests.
+# Every other module tests internal helpers against the real API.
+# Exception here: the bugs fixed in melodi/data/py live inside parsing helpers
+# called inside get_dataset. Testing them through the real API would
+# require downloading DS_IPC_PRINC on every test run.
```
tgrandje, line 18: Thanks for the explanations! As `pynsee` was never (yet) released with MELODI api, what do you think of removing the comments detailing the tests' purposes as fixes to my original branch? In case you're ok, I'll flag the involved comments I've seen.

tests/melodi/test_pynsee_melodi.py
```diff
@@ -0,0 +1,261 @@
...
+        test = test & ("attributes" not in df.columns)
+        self.assertTrue(test)
+
+    def test_get_catalog(self):
+        df = get_catalog()
+        self.assertTrue(isinstance(df, pd.DataFrame))
+
+    def test_get_catalog_known_datasets_present(self):
+        # dataset_identifier is not unique (one dataset can have multiple products).
+        # Assert presence of stable, long-running datasets rather than len > 0.
```
tgrandje, line 166: It took me a while to understand what you meant with the "rather tan len > 0". Maybe we could also remove that part, as it doesn't bring anything useful (I think): I let you be the judge on that 😉 ! (Note: it might very well just me being daft...)

tests/melodi/test_pynsee_melodi.py
```diff
@@ -0,0 +1,261 @@
...
+    def test_get_range_sparse(self):
+        # Assert exact dimension set, not just column presence.
+        test = True
+        df = get_range("DS_TICM_PRATIQUES", language="fr")
+        test = test & isinstance(df, pd.DataFrame)
+        test = test & (set(df["concept_code"]) == _TICM_CONCEPT_CODES)
+        self.assertTrue(test)
+
+    def test_get_range_with_values(self):
+        # Assert exact measure set for TICM_MEASURE.
```
tgrandje, line 183: Maybe we could bring here the comment of _TICM_MEASURES (as it being a test of a stable structure in the INSEE's side?).

tests/melodi/test_pynsee_melodi.py
```diff
@@ -0,0 +1,261 @@
...
+        # Assert presence of stable, long-running datasets rather than len > 0.
+        test = True
+        df = get_catalog(language="fr")
+        known = {"DS_IPC_PRINC", "DS_TICM_PRATIQUES", "DS_ICA"}
+        test = test & isinstance(df, pd.DataFrame)
+        test = test & known.issubset(set(df["dataset_identifier"]))
+        self.assertTrue(test)
+
+    def test_get_range_sparse(self):
+        # Assert exact dimension set, not just column presence.
```
tgrandje, line 175: Maybe we could bring here the comment of _TICM_CONCEPT_CODES (as it being a test of a stable structure in the INSEE's side?).

tests/melodi/test_pynsee_melodi.py
```diff
@@ -0,0 +1,261 @@
...
+        df = get_dataset(
+            "DS_TICM_PRATIQUES",
+            language="fr",
+            time_period=2025,
+            sex="F",
+            age="Y45T59",
+        )
+        test = isinstance(df, pd.DataFrame)
+        test = test & (len(df) == len(_TICM_MEASURES))
+        test = test & (set(df["TICM_MEASURE"]) == _TICM_MEASURES)
```
tgrandje, line 206: I kind of feel that we're testing the same thing here and at test_get_range_with_values (in case of failure because of INSEE's updates, we should get 2 errors and I'm not sure we will could do much altogether). I'm thinking we might instead check if the dimensions are coherent with those of the `get_range` function, what do you think?

tests/melodi/test_pynsee_melodi.py
```diff
@@ -0,0 +1,261 @@
...
+        # language="fr" must exclude _en columns
+        meta = _parse_metadata(
+            _mock_response(_METADATA_PAYLOAD), language="fr"
+        )
+        test = ("title_fr" in meta) & ("title_en" not in meta)
+        test = test & ("publisher_fr" in meta) & ("publisher_en" not in meta)
+        self.assertTrue(test)
+
+    def test_parse_observations_dynamic_measure_key(self):
+        # Regression: measure key must be extracted dynamically, not hardcoded.
```
tgrandje, line 135: Update commentary in context with pynsee's release with brand new MELODI api? (maybe just remove the regression part?)

tests/melodi/test_pynsee_melodi.py
```diff
@@ -0,0 +1,261 @@
...
+        df = _parse_dataset_observations(_mock_response(_OBS_DYNAMIC_MEASURE))
+        test = isinstance(df, pd.DataFrame)
+        test = test & ("OBS_VALUE_INDICE_DE_PRIX" in df.columns)
+        test = test & ("OBS_VALUE_NIVEAU" not in df.columns)
+        test = test & (df["OBS_VALUE_INDICE_DE_PRIX"].iloc[0] == 105.3)
+        test = test & pd.isna(df["OBS_VALUE_INDICE_DE_PRIX"].iloc[1])
+        self.assertTrue(test)
+
+    def test_parse_observations_no_measures_column(self):
+        # Regression: observations without "measures" must not crash.
```
tgrandje, line 145: Update commentary in context with pynsee's release with brand new MELODI api? (maybe just remove the regression part?)

tests/melodi/test_pynsee_melodi.py
```diff
@@ -0,0 +1,261 @@
...
+        test = test & df["OBS_VALUE_INDICE_DE_PRIX"].notna().any()
+        self.assertTrue(test)
+
+    def test_get_idbank_1(self):
+        df = get_idbank("010770930")
+        self.assertTrue(isinstance(df, pd.DataFrame))
+
+    def test_get_idbank_series_history_stable(self):
+        # 010770930 = DS_ICA monthly series, sector 46.19A (non-food purchasing
+        # centres), base 2021. Open series: do NOT assert len (grows monthly).
```
tgrandje, line 228: I get it we shouldn't check the len, but that does'nt mean we couldn't check wether it's still updated. I was thinking of something like `(df.groupby(df["TIME_PERIOD"].str[:4])["dataset"].count()[:-1]==12).all()`, what do you think?

tests/melodi/test_pynsee_melodi.py
```diff
@@ -0,0 +1,261 @@
...
+    def test_get_idbank_multi(self):
+        # nunique == 2 confirms both series were returned, not one duplicated.
+        df = get_idbank("010598544+010770930")
+        test = isinstance(df, pd.DataFrame)
+        test = test & (df["idBank"].nunique() == 2)
+        self.assertTrue(test)
+
+    def test_get_idbank_language_fr(self):
+        # Regression: "key2 in {'all', language}" was backwards: _fr columns
+        # were absent when language="fr". Fixed to "language in {'all', key2}".
```
tgrandje, line 246: Update commentary in context with pynsee's release with brand new MELODI api?

tests/melodi/test_pynsee_melodi.py
```diff
@@ -0,0 +1,261 @@
...
+        df = get_idbank("010770930", language="fr")
+        test = isinstance(df, pd.DataFrame)
+        test = test & any(c.endswith("_fr") for c in df.columns)
+        self.assertTrue(test)
+
+    def test_get_idbank_dead_bdm_ids(self):
+        # BDM series with no MELODI dataset yet -> HTTP 200 + empty list -> empty df.
+        df = get_idbank("001565530+001565531")
```
tgrandje, line 254: I'm thinking this would be better yet to do something quite explicit with a `df = get_idbank("spam+eggs")`; for those of us outside INSEE, it's not the most convenient of things to retrieve idbanks and check if they have associated data 😉 . By the way, should we log a warning stating that no data was retrieved and that is not a pynsee error?

docs/modules.rst
```diff
@@ -48,3 +48,9 @@ Utils
 
 .. automodule:: pynsee.utils
     :members:
+
+MELODI
+------
+
+.. automodule:: pynsee.melodi
+    :members:
```
tgrandje, line 56: What do you think of putting MELODI before utils? I'd even go as far as setting it where METADATA was (and setting METADATA at the latest position in regard of it's deprecation).

tgrandje:
Hi @GrasClement , sorry I took so long to review this, I was overloaded at work.

So I finally took the time to do it, with the exception of testing the documentation (I'm in a train on a personnal windows machine...) though it looks ok to me. I've made some comments: depending of your available time, feel free to let me know if you haven't the time to check all of this.

In any case, thank you for both your great work and the effort to uniformize the API with the silent/update kwargs ❤️ !

GrasClement:
Hi @tgrandje, no worries about the timing, and thanks for taking the time to review the PR!

I don't seem to see any review comments or additional commits attached to the PR yet. Whenever you have a chance, could you check whether the comments were actually submitted? Or maybe the issue is on my side and I'm looking in the wrong place.

I'd be happy to go through your feedback and update the PR once this is sorted out 😎

tgrandje (changes requested):
Sorry @GrasClement I forgot to submit the review 😳 !

pynsee/melodi/data.py
GrasClement: Done, translated to English.

pynsee/melodi/data.py
GrasClement: You're right, I dropped the apply.

pynsee/melodi/data.py
GrasClement: Removed

tests/melodi/test_pynsee_melodi.py
GrasClement: Reworded that one, and two more with the same "fix" framing I noticed, figured I'd just fix those I found rather than wait for you to flag them separately.

tests/melodi/test_pynsee_melodi.py
GrasClement: Removed :)

tests/melodi/test_pynsee_melodi.py
GrasClement: Done

tests/melodi/test_pynsee_melodi.py
GrasClement: Done

tests/melodi/test_pynsee_melodi.py
GrasClement: This one was a bit harder but I think I got your point as: make get_dataset's check coherent with get_range instead of a second hardcoded comparison. But I figured  that could lead to a silent-pass problem if something affects both functions identically. So I dropped the measure-set check here since it's covered in test_get_range_with_values, and added a check for what's  unique to get_dataset: no duplicate rows per measure, and confirmation the filters were applied. Let me know if that's not what you had in mind.

tests/melodi/test_pynsee_melodi.py
GrasClement: Done

tests/melodi/test_pynsee_melodi.py
GrasClement: Done

tests/melodi/test_pynsee_melodi.py
GrasClement: You're right, dataset is there and it's never null for this series. I went with your version.

tests/melodi/test_pynsee_melodi.py
GrasClement: Done

tests/melodi/test_pynsee_melodi.py
GrasClement: Yup, the API returns 200 + an empty list for literally any unrecognized idbank, not just real dead-BDM ones, so spam+eggs works the same and reads better. Switched to that. Also added the warning in data.py : get_idbank logs now when there's no data :)

docs/modules.rst
GrasClement: Done

GrasClement:
Hi @tgrandje 

Replied to all your comments, thanks again for the thorough review, and sorry it took me a while to get back to it. 

Also found and fixed one unrelated thing while I was in there: the get_range docstring example was out of date. It might be because of an API change INSEE pushed in July, nothing to do with your comments.

Let me know if anything needs another pass. I tried to finish cleanly to let new contributors jump in (see new comments in #307).

Cheers :)

